PR #6, opened 2021-12-12
options: Report image format and virtual size

nirs:
Report image virtual size in OPTIONS so clients can get the image size
without a possibly slow extent call.

Report the format since OPTIONS can report the virtual size only when
the backend provides raw format. This is used when using the http
backend to report OPTIONS to the client.

Reporting virtual size is easy with the nbd and memory backends since
they always use raw format.

When using file backend and qcow2 image, we don't have access to the
image virtual size, and this size is not helpful to the user uploading
or downloading data. Currently we don't know about the image format
since engine does not report it in the ticket.

The http backend reports the info from the remote server, so it depends
on the backend used by the remote server, and on having new server
reporting the format and size.

To keep code and the API simple, we report virtual size only when using
the nbd and memory backends. When engine will report the image format
for the file backend, we can also report the size for raw images access
via the file backend.

This is work in progress, exploring the needed changes. I will post a
cleaner version later.


docs/images.md
```diff
@@ -84,6 +84,25 @@ When using multiple writers, each writer should modify a distinct byte
 range. If two writers modify the same byte range concurrently they will
 overwrite each other data.
 
+### format
```
nirs, line 87: Maybe this should be transfer_format to make it more clear. It is not the same as image format - when using the nbd backend, we always transfer raw data, even if the underlying image is using qcow2 format.

docs/images.md
```diff
@@ -112,9 +131,9 @@ Response:
     Content-Length: 122
 
     {"unix_socket": "\u0000/org/ovirt/imageio", "features": ["extents", "zero", "flush"],
-     "max_readers": 8, "max_writers": 8}
+     "max_readers": 8, "max_writers": 8, "format": "raw", "size": 53687091200}
```
nirs, line 134: The docs discuss `virtual_size` but here we report `size`. I think reporting `virtaul_size` will be more clear. This is also the same term used in `qemu-img info`.

docs/images.md
```diff
@@ -125,11 +144,13 @@ Get options for ticket-id with read-write access using nbd backend:
         "flush"
       ],
       "max_readers": 8,
-      "max_writers": 8
+      "max_writers": 8,
+      "format": "raw",
+      "virtual_size": 53687091200
```
nirs, line 149: Here we report `virtual_size` so the previous `size` is likey an error.

docs/images.md
```diff
@@ -143,6 +164,9 @@ Get options for ticket-id with read-only access using file backend:
       "max_writers": 1
     }
 
+Note that `size` and `format` are not reported, and this backend does
+not support multiple writers.
```
nirs, line 168: I think this is goo enough for the legacy file backend. We don't want people to use it and we better not spend time on making it better. New features should be available only with the nbd backend.

ovirt_imageio/_internal/backends/http.py
```diff
@@ -325,10 +340,17 @@ def seek(self, n, how=os.SEEK_SET):
...
 
     def size(self):
-        # We have 2 bad options:
-        # - Get last extent, may be slow, and may not be neded otherwise.
-        # - Emulate HEAD request, logging tracebacks in the remote server.
-        # Getting extents is more polite, so lets use it if we can.
+        """
+        Return backend size in bytes.
+
+        With newer server (imageio >= 2.3) reporting the virtual size the size
```
nirs, line 346: The version here should probably be  >= 2.5 - the version including this feature.

test/handlers/images_test.py
```diff
@@ -750,8 +759,110 @@ def test_options_write(srv, client, tmpdir):
...
+    KiB,
+    512 * KiB,
+    MiB,
+    512 * MiB,
+    GiB,
+    512 * GiB,
+    TiB,
+    8 * TiB
+])
+def test_options_nbd_size_check(srv, client, tmpdir, nbd_server, size):
```
nirs, line 845: I think this is already covered by the nbd_{read,write} tests, and testing different sizes do not add enough value to run this test multiple times.

aesteve-rh:
@nirs I have done some tests and can see the new options available when I run an example server locally, with a nbd backend. It looks correct.

But unfortunately I haven't been able to do any test showing any relevant speedup. It downloads at roughly the same speed as master branch. Tried two methods: with direct download with a GET request, and through the client using the `imageio-client` in the examples folder. Also tried with empty 8TiB images, and with fedora-36 image, resized to 8 TiB.

I would like to have some data to share before ready this PR up, but I'm not sure what else can I try to obtain it, or what I may be missing in my own local testing.

nirs:
@aesteve-rh with download you will not see any difference, since we always get extents once.

You should see a difference with incremental backup. With current release we get extent twice,
once with context=zero to get the size of the source, and then with context=dirty to get dirty blocks information. With 8TiB image the extents call takes 24 seconds (on my system) so calling twice will spend 48 seconds only on the extents calls. With the new feature we will call extents only once, since image size is available via OPTIONS.

nirs:
Let’s close since it stale for long time.

PR #18, opened 2022-01-15
docker: Update centos containers repos

nirs:
Install ovirt-release-master from copr instead of stale
resources.ovirt.org.

### DIFF
--- a/docker/Dockerfile.centos-8
+++ b/docker/Dockerfile.centos-8
@@ -1,8 +1,11 @@
 FROM quay.io/centos/centos:stream8
 
 # Runtime packages.
-RUN dnf install -y \
-        http://resources.ovirt.org/pub/yum-repo/ovirt-release-master.rpm \
+RUN dnf install -y dnf-plugins-core \
+    && dnf copr enable -y \
+        ovirt/ovirt-master-snapshot \
+        centos-stream-8 \
+    && dnf install -y ovirt-release-master \
     && dnf install -y \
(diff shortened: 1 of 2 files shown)


vjuranek (approved):

vjuranek:
This branch is out-of-date with the base branch. Not sure why rebase and merge cannot be applied, but I guess `Update branch` button will create merge commit and will break nice linear history ... Or do you don't mind merge commits (personally I don't like them much)?

nirs:
> This branch is out-of-date with the base branch. Not sure why rebase and merge cannot be applied,

Because I enabled the branch protection rule:

    Require branches to be up to date before merging

This ensures that the tests pass with a rebased patch before merging.
The author of the PR needs to rebase it before merge.

I added you to the list of actors that can bypass the requirement, so your should
be able to rebase and merge now.

> but I guess `Update branch` button will create merge commit and will break nice linear history ... Or do you don't mind merge commits (personally I don't like them much)?

I'm not sure what will happen when using "Update branch" but I don't like to have
merge commits if we can avoid them.

nirs:
Test failure on centos 9 is a bug in qemu 6.2.0, reporoduced also on
Fedora 35.

Once the new containers are built and we get qemu 6.2.0 in all
containers, we expect all the builds to fail.

Reported the issue here:
https://lists.nongnu.org/archive/html/qemu-block/2022-01/msg00292.html

Issue #13, opened 2021-12-14
Build containers every week


nirs:
Instead of hacking quay.io to build new container by changing the container file source,
build with podman without cache, and push to quay.io.

The main issue is how to manage permissions to push to quay.io. This pull
request show how to do this:
https://github.com/oVirt/ovirt-provider-ovn/pull/5

PR #19, opened 2022-01-16
docker: Disable quay.io caching

nirs:
quay.io never updates our containers unless we change the list of
packages, so we don't get updates when new packages are released, and we
fail to detect regressions in packages we depend on.

Add a version to the RUN line. The version should be modified each time
we want to rebuild the containers.

Adding a workflow the build the container is probably better, but
require more work.

### DIFF
--- a/docker/Dockerfile.centos-8
+++ b/docker/Dockerfile.centos-8
@@ -1,7 +1,8 @@
 FROM quay.io/centos/centos:stream8
 
 # Runtime packages.
-RUN dnf install -y dnf-plugins-core \
+RUN echo v1 \
+    && dnf install -y dnf-plugins-core \
     && dnf copr enable -y \
         ovirt/ovirt-master-snapshot \
         centos-stream-8 \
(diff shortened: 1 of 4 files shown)


nirs:
The failing tests on centos 9 are know qemu-nbd issue:
https://lists.nongnu.org/archive/html/qemu-block/2022-01/msg00292.html

vjuranek (commented):
Usually containers for all platforms are built, which requires changing version on 4 places. It would be better to have a version file which we would contain the version can `cat` it in the Dockerfiles. Or it doesn't work and we have to really modify the Dockerfile?

nirs:
> Usually containers for all platforms are built, which requires changing version on 4 places. It would be better to have a version file which we would contain the version can `cat` it in the Dockerfiles. Or it doesn't work and we have to really modify the Dockerfile?

If we want to disable the cache we need to modify the RUN command. If we
spend time on smarter solution (e.g. genrate the docker files) it is better
to spend the time on an action building the containers in github and pushing
them to quay, see #13.

I posted this as a quick solution for now.

PR #28, opened 2022-01-30
ovirt-img: Improve error handling and documention

nirs:
- Exit with clear error instead of panic if extent callback failed
- Improve comment about EINTR in blockStatus()

### DIFF
--- a/ovirt-img/nbd/nbd.go
+++ b/ovirt-img/nbd/nbd.go
@@ -151,8 +151,9 @@ func (b *Backend) blockStatus(offset, length uint64) ([]uint32, error) {
 		return 0
 	}
 
-	// BlockStatus may fail randomly, looks like bug in libnbd.
-	// https://listman.redhat.com/archives/libguestfs/2021-October/msg00113.html
+	// BlockStatus may fail randomly with EINTR. Fixed in libnbd 1.10.2
+	// https://gitlab.com/nbdkit/libnbd/-/commit/48136328150bc587178091b5766bda382158cb6c
+	// TODO remove when we require libnbd >= 1.10.2.
 	for {
 		err := b.h.BlockStatus(length, offset, cb, nil)
 		if err == nil {
--- a/ovirt-img/nbd/nbd.go
+++ b/ovirt-img/nbd/nbd.go
@@ -144,10 +144,8 @@ func (b *Backend) blockStatus(offset, length uint64) ([]uint32, error) {
 	var result []uint32
 
 	cb := func(metacontext string, offset uint64, e []uint32, error *int) int {
-		if *error != 0 {
-			panic("expected *error == 0")
-		}
-		if metacontext == "base:allocation" {
+		// blockStatus() will fail on non zero *error.
+		if metacontext == "base:allocation" && *error == 0 {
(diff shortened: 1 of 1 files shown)


bennyz (approved):

ovirt-img/nbd/nbd.go
```diff
@@ -144,17 +144,16 @@ func (b *Backend) blockStatus(offset, length uint64) ([]uint32, error) {
...
 			result = e
 		}
 		return 0
 	}
 
-	// BlockStatus may fail randomly, looks like bug in libnbd.
-	// https://listman.redhat.com/archives/libguestfs/2021-October/msg00113.html
+	// BlockStatus may fail randomly with EINTR. Fixed in libnbd 1.10.2
+	// https://gitlab.com/nbdkit/libnbd/-/commit/48136328150bc587178091b5766bda382158cb6c
+	// TODO remove when we require libnbd >= 1.10.2.
```
barpavel, line 156: TODO
-->
TODO:

ovirt-img/nbd/nbd.go
```diff
@@ -144,17 +144,16 @@ func (b *Backend) blockStatus(offset, length uint64) ([]uint32, error) {
 	var result []uint32
 
 	cb := func(metacontext string, offset uint64, e []uint32, error *int) int {
-		if *error != 0 {
-			panic("expected *error == 0")
-		}
-		if metacontext == "base:allocation" {
+		// blockStatus() will fail on non zero *error.
```
barpavel, line 147: Maybe I'm missing some Go knowledge - what do you mean by this comment "blockStatus() will fail on non zero *error." and by PR comment "Exit with clear error instead of panic"?
If *error == 0, then local variable "result" is actually undefined/nil?
That is what you mean by "will fail" - will return uninitialized value? And where is "Exit with clear error" part? :)
Again, maybe I'm missing Go language syntax knowledge :)

vjuranek: > what do you mean by this comment "blockStatus() will fail on non zero *error." and by PR comment "Exit with clear error instead of panic"?

See few lines bellow (which probably answers your other questions):

```
if err.(*libnbd.LibnbdError).Errno != syscall.EINTR {
			return nil, err
}
```

ovirt-img/nbd/nbd.go
vjuranek: not sure if worth fixing

vjuranek (approved):

ovirt-img/nbd/nbd.go
nirs: This is not about Go but about how libnbd works.

When you call blockStatus() you give it a callback (this is our cb). This callback is called once for every meta context. Since we have register only one ("base:allocation") it will be called once.

If the server returned an error, this callback will be called with *error != 0. We can use this info to avoid coping the entries on errors.

When *error != 0, the call to blockStatus() fails and return a libnbd.LibnbdError object, including the error code and error message. Our function wil fail with this error.

So we don't need to panic() in the callback as we did before, it is not helpful and it hides the nice error message returned by blockStatus().

PR #30, opened 2022-02-04
Speed up ranges calculations

### LINKED ISSUE #29: Inefficient measuring of transferred size
Each time we finish an operation, we add the operation range
to a sorted completed range list, a merge the result:

    # ovirt_imageio/_internal/auth.py
    r = measure.Range(op.offset, op.offset + op.done)
    bisect.insort(self._completed, r)
    self._completed = measure.merge_ranges(self._completed)

This is not optimal since measure.merge sort the sorted list:

    # ovirt_imageio/_internal/measure.py
    def merge_ranges(ranges):
        ranges = sorted(ranges)
        ...

So we pay twice for sorting - once when inserting, second time when merging
instead of only once. This can harm uploads using very small requests. Example
use case is older virt-v2v.

Even worse, when reporting transferred size we copy the completed
range, sort with *unordered* ongoing operations, and merge (sorting again).

    def transferred(self):
        """
        The number of bytes that were transferred so far using this ticket.
        """
        if len(self.ops) > 1:
            # Both read and write, cannot report meaningful value.
            return None

        with self._lock:
            # NOTE: this must not modify the ticket state.
            completed = [measure.Range(r.start, r.end)
                         for r in self._completed]
            ongoing = [measure.Range(op.offset, op.offset + op.done)
                       for op in self._ongoing]

        ranges = sorted(completed + ongoing)
        ranges = measure.merge_ranges(ranges)
        return sum(len(r) for r in ranges)

So we sort twice when inserting, and twice when reporting the value.

The related benchmarks do not look too bad:

Running operation (called many time per second):

    test/auth_test.py::test_run_operation_benchmark 100000 operations, 4 concurrent threads in 1.990 seconds (19899 nsec/op)

Computing transferred bytes (called once per 10 seconds)

    test/auth_test.py::test_transferred_benchmark[1] 1 concurrent operations, 10000 calls in 0.037 seconds (3677 nsec/op)
    test/auth_test.py::test_transferred_benchmark[2] 2 concurrent operations, 10000 calls in 0.058 seconds (5809 nsec/op)
    test/auth_test.py::test_transferred_benchmark[4] 4 concurrent operations, 10000 calls in 0.095 seconds (9549 nsec/op)
    test/auth_test.py::test_transferred_benchmark[8] 8 concurrent operations, 10000 calls in 0.180 seconds (18011 nsec/op)

transfered() is called once per 10 seconds when engine monitored
ticket status. Adding completed operations is called up many times
per seconds (depending on size of requests).

We report the ongoing operations since a transfer can include one
operation covering the entire image. If we won't report ongoing
operations, we will have no progress for such transfers.

Operations are kept in a set, to make adding and removing operations
efficient. With advanced client we have multiple connections and up
to 8 concurrent operations.

We can fix this by keeping sorted list of completed operations, copying
it and inserting the ongoing operations. Since we have maximum 8 ongoing
operations, this should be O(logN) operation.

The code to manage a sorted range list should be encapsulated in a class,
something like:

    completed = RangeList()

    r = Range(start, end)
    completed.add(r)

    transferred = RangeList(completed)
    
    for op in onging:
        r = Range(op.start, op.start + op.done)
        transferred.add(r)

    total = transferred.sum()

nirs:
Speed up ranges calculations when completing requests and computing transferred size.

- Make it easier to run the benchmarks
- Improve the benchmarks simulating real transfer with 2 kinds of clients (nbdcopy, imageio)
- Improve benchmarks output
- Improve measure tests
- Remove python 2 only `Range.__ne__`
- Add `Range.__repr__` for easier debugging
- Add `Range.__slots__` to minimize Range creation time
- Speed up `Range.__lt__`
- Remove unneeded sorting during merge
- Add `RangeList` class for managing ranges

The actual speedup is very small, but the code is cleaner and easier to maintain.

Fixes #29.

### DIFF
--- a/ovirt_imageio/_internal/measure.py
+++ b/ovirt_imageio/_internal/measure.py
@@ -16,7 +16,11 @@ class Range:
         self.end = end
 
     def __lt__(self, other):
-        return (self.start, self.end) < (other.start, other.end)
+        if self.start < other.start:
+            return True
+        if self.start == other.start:
+            return self.end < other.end
+        return False
 
     def __len__(self):
         return self.end - self.start
--- a/ovirt_imageio/_internal/measure.py
+++ b/ovirt_imageio/_internal/measure.py
@@ -6,6 +6,8 @@
 # the Free Software Foundation; either version 2 of the License, or
 # (at your option) any later version.
 
+import bisect
+
 
 class Range:
 
@@ -34,19 +36,61 @@ class Range:
         return f"Range(start={self.start}, end={self.end})"
 
 
-def merge_ranges(ranges):
-    """
-    Gets an iterable of Range objects and returns
-    a sorted list of the merged ranges.
-    """
-    ranges = sorted(ranges)
-    if not ranges:
-        return ranges
-    merged = [ranges.pop(0)]
-    for range in ranges:
-        current = merged[-1]
-        if current.end >= range.start:
-            current.end = max(current.end, range.end)
+class RangeList:
+
+    def __init__(self, other=None):
+        if other:
+            self._ranges = [Range(r.start, r.end) for r in other._ranges]
         else:
-            merged.append(range)
+            self._ranges = []
+
+    def add(self, r):
+        """
+        Add a single range.
+        """
+        bisect.insort_left(self._ranges, r)
+        self._ranges = _merged(self._ranges)
+
+    def update(self, rs):
+        """
+        Update from iterable of unsorted ranges.
+        """
+        self._ranges.extend(rs)
+        self._ranges.sort()
+        self._ranges = _merged(self._ranges)
+
+    def sum(self):
+        return sum(len(r) for r in self._ranges)
+
+
+def _merged(ranges):
+    """
+    Merge sorted list of ranges.
+
+    The ranges are sorted, but may contain:
+
+    - consecutive ranges (very likely):
+
+      [Range(0, 100), Range(100, 200)]
+
+    - duplicate ranges (unlikely):
+
+      [Range(0, 100), Range(0, 100)]
+
+    - ranges overlapping other ranges (unlikely):
+
+      [Range(0, 100), Range(50, 200)]
+
+    Return a merged list of ranges without duplicates, overlaps, or
+    consecutive ranges.
+    """
+    merged = ranges[:1]
+
+    for r in ranges[1:]:
+        current = merged[-1]
+        if current.end >= r.start:
+            current.end = max(current.end, r.end)
+        else:
+            merged.append(r)
+
     return merged
--- a/ovirt_imageio/_internal/measure.py
+++ b/ovirt_imageio/_internal/measure.py
@@ -9,6 +9,8 @@
 
 class Range:
 
+    __slots__ = ("start", "end")
+
     def __init__(self, start, end):
         self.start = start
         self.end = end
@@ -24,8 +26,8 @@ class Range:
                 self.start == other.start and
                 self.end == other.end)
 
-    def __ne__(self, other):
-        return not self == other
+    def __repr__(self):
+        return f"Range(start={self.start}, end={self.end})"
 
 
 def merge_ranges(ranges):
--- a/test/auth_test.py
+++ b/test/auth_test.py
@@ -6,6 +6,8 @@
 # the Free Software Foundation; either version 2 of the License, or
 # (at your option) any later version.
 
+import itertools
+import logging
 import time
 
 import pytest
@@ -19,6 +21,8 @@ from ovirt_imageio._internal.auth import Ticket, Authorizer
 from test import testutil
 
 CHUNK_SIZE = 8 * 1024**2
+
+log = logging.getLogger("test")
 
 
 class Context:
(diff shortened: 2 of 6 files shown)


test/auth_test.py
```diff
@@ -209,52 +213,73 @@ def test_transfered_ongoing_non_continues_ops(cfg):
 
 
 @pytest.mark.benchmark
-def test_run_operation_benchmark(cfg):
-    # Run 1000000 operations with 4 concurrent threads.
+@pytest.mark.parametrize("workers", [1, 2, 4, 8])
+def test_run_operation_benchmark(cfg, workers):
+    # Simulate tranferr of a big image using nbdcopy as a client. We split the
```
bennyz, line 218:
```suggestion
    # Simulate tranfer of a big image using nbdcopy as a client. We split the
```

test/auth_test.py
```diff
@@ -209,52 +213,73 @@ def test_transfered_ongoing_non_continues_ops(cfg):
...
+    # image to 128 MiB segments. Every wroker send one segment and get the next
```
bennyz, line 219:
```suggestion
    # image to 128 MiB segments. Every worker sends one segment and gets the next
```

test/auth_test.py
barpavel: tranferr --> tranfer --> transfer :)

test/auth_test.py
```diff
@@ -209,52 +213,73 @@ def test_transfered_ongoing_non_continues_ops(cfg):
...
-    operations = 10**6
-    workers = 4
-    chunk = 10**9
-    step = chunk * workers // operations
 
-    def worker(offset, size):
-        while offset < size:
-            ticket.run(Operation(offset, step))
-            offset += step
+    image_size = 100 * 1024**3
```
barpavel, line 224: Maybe it was worth to add "MiB" & "GiB" constants similar to "nbdutil_test.py".

test/auth_test.py
```diff
@@ -209,52 +213,73 @@ def test_transfered_ongoing_non_continues_ops(cfg):
...
         for i in range(workers):
-            t = util.start_thread(worker, args=(i * chunk, chunk))
+            t = util.start_thread(worker, name=f"run/{i}")
             threads.append(t)
     finally:
         for t in threads:
             t.join()
 
     elapsed = time.monotonic() - start
+    iops = image_size / io_size
```
barpavel, line 261: Again, wouldn't you like to have an integer here, so need to use "//"?

test/auth_test.py
```diff
@@ -209,52 +213,73 @@ def test_transfered_ongoing_non_continues_ops(cfg):
...
 
-    def worker(offset, size):
-        while offset < size:
-            ticket.run(Operation(offset, step))
-            offset += step
+    image_size = 100 * 1024**3
+    segment_size = 128 * 1024**2
+    io_size = 2 * 1024**2
+
+    segment_count = (image_size + segment_size - 1) / segment_size
```
barpavel, line 228: I'm vaguely starting to recall the Python 3 porting work :)
Wouldn't that result a "float"? Thus you need to use "//".
Or even better use:
math.ceil(image_size/segment_size)

test/auth_test.py
nirs: Thanks, will update :-)

test/auth_test.py
nirs: Good idea, I opened #31 for this.

test/auth_test.py
nirs: Yes, this is a bug, should be //. It works because I chose size which
is multiple of io_size.

test/auth_test.py
nirs: Yes, should be fixed.

test/auth_test.py
```diff
@@ -6,6 +6,8 @@
 # the Free Software Foundation; either version 2 of the License, or
```
vjuranek, line 6: last sentence of commit messages has double not: "(5%) is not not that"

test/measure_test.py
```diff
@@ -11,83 +11,25 @@
 from ovirt_imageio._internal import measure
```
vjuranek: "it's" -> "its" in "if it's start start value is smaller"

ovirt_imageio/_internal/measure.py
```diff
@@ -34,19 +36,61 @@ def __repr__(self):
...
+
+    - duplicate ranges
+
+      [Range(0, 100), Range(0, 100)]
+
+    - ranges overlapping other ranges around them
+
+      [Range(0, 100), Range(50, 200)]
+
+    - concecutive ranges:
```
vjuranek, line 80: concecutive -> consecutive

test/auth_test.py
```diff
@@ -208,35 +212,133 @@ def test_transfered_ongoing_non_continues_ops(cfg):
...
+    def run(self):
+        start = time.monotonic()
+        threads = []
+        try:
+            for i in range(self.workers):
+                t = util.start_thread(self._worker, name=f"{self.name}/{i}")
+                threads.append(t)
+        finally:
+            for t in threads:
+                t.join()
```
vjuranek, line 229: maybe add some timeout?

test/auth_test.py
```diff
@@ -208,35 +212,133 @@ def test_transfered_ongoing_non_continues_ops(cfg):
...
-            t = util.start_thread(worker, args=(i * chunk, chunk))
-            threads.append(t)
-    finally:
-        for t in threads:
-            t.join()
+    The image is split to 128 MiB segments, and each threads send one segment
+    at a time. Request size is typically small.
 
-    elapsed = time.monotonic() - start
+    Because very worker is sending to a different area, Completed ranges are
```
vjuranek, line 243: "Completed" -> "completed" (lower "c")

test/auth_test.py
```diff
@@ -208,35 +212,133 @@ def test_transfered_ongoing_non_continues_ops(cfg):
...
+        n = next(self.segment_counter)
+        if n < self.segment_count:
+            return n
+
+
+class Imageio(Client):
+    """
+    Simulate imageio client calls.
+
+    Workers threads read extents from a queue and send requests. Requests are
```
vjuranek, line 286: "Workers" -> "Worker"

vjuranek (changes requested):

test/auth_test.py
nirs: I don't think it worth the effort. The benchmarks do not run in the CI, and when you run them locally you can easily interrupt the run in needed. What can make sens i using the pytest-timeout plugin.

I think the next step is to enable the benchmarks in the CI, to make sure they do not
break, and enable pytest timeout for the benchmarks.

nirs: Turns out we already enable pytest-timeout with 30 seconds timeout per test, so the only thing missing is to enable the benchmarks also in the CI.

We can create special configuration with small number of iterations just to keep the benchmarks running, but I'm not sure it worth the effort.

vjuranek (approved):

PR #38, opened 2022-02-20
Zero detection in client side

### LINKED ISSUE #37: Detect zeroes during uploads on client side
Currently we detect zeroes and sparsify on the server side. This is the only way to
support stupid clients like browsers, used when uploading images via the webadmin portal.

When uploading images using the API, client can save lot of bandwidth by detecting zeroes
on the client side and sending a tiny zero request instead of sending actual zeroes over
the wire.

On the other hand, detecting zeroes does not works with the smart http code trying to minimize
the number of http request, so I'm not sure this will improve the performance in most cases.

This requires testing, and may be added as a optional features that will be applied only in
certain cases.

nirs:
This is experimental work on zero detection in client side. Some images
may contain zeroes in data extents. Currently we send these extents as
data, and qemu-nbd detect zeroes and convert them to hole or allocated
area.

Detecting zeroes on client side is much more efficient; instead of sending
zeroes over the wire, we send a tiny zero request without any payload.
Practically this is very hard to do while using the http optimized code that
try to minimize the number of http requests.

I implemented zero detection in a very simplistic way. Testing show small
improvement in the best case, fully allocated image full of zeroes, and
large slowdown when uploading real images.

Zero detection can be useful when uploading over very slow network. I did
not test this case yet.

Unfinished:
- Add upload benchmarks
- Should we adapt buffer size automatically if --detect-zeroes is set, and
  --buffer-size was not set?
- Is is possible to implement in a better way by buffering more data in the
  copy loop?

Fixes #37

### DIFF
--- a/bench/conf.d/daemon.conf
+++ b/bench/conf.d/daemon.conf
@@ -1,0 +1,28 @@
+# Configuration for benchmarking the server.
+#
+#    ./ovirt-imageio -c bench
+
+[backend_nbd]
+buffer_size = 262144
+#buffer_size = 524288
+#buffer_size = 1048576
+#buffer_size = 2097152
+#buffer_size = 4194304
+
+[tls]
+key_file = test/pki/system/key.pem
+cert_file = test/pki/system/cert.pem
+ca_file = test/pki/system/ca.pem
+
+[control]
+transport = unix
+socket = bench/daemon.sock
+
+[logger_root]
+level = INFO
+
+[handler_logfile]
+args = ("bench/daemon.log",)
+
+[profile]
+filename = bench/daemon.prof
--- a/bench/upload.json
+++ b/bench/upload.json
@@ -1,0 +1,9 @@
+{
+    "uuid": "upload",
+    "size": 6442450944,
+    "url": "nbd:unix:/tmp/nbd.sock",
+    "timeout": 3000,
+    "inactivity_timeout": 300,
+    "sparse": true,
+    "ops": ["write"]
+}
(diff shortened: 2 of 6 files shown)


vjuranek (commented):
Tha changes LGTM, nut I'd expect better numbers at least for `zero-6g.raw` case. Maybe it because it's local copy from one drive to another? It would be nice to see some numbers from tst when client and server are different hosts.

nirs:
> Tha changes LGTM, nut I'd expect better numbers at least for `zero-6g.raw` case. Maybe it because it's local copy from one drive to another? It would be nice to see some numbers from tst when client and server are different hosts.

Coping from local drive to another local drive is usually the fastest way you can
copy, but in this case we don't write anything to the target disk since the entire
image is zero.

In the zero case, we detect that all the data we read from qemu-nbd is zero, and we
send one zero request for every read to imageio. Processing zero request in imageio
with the storage tested (NFS) is extremely fast (~60 GiB/s).

Basically this what the test show that it takes 3.8 seconds to read and detect 6 GiB
of zeroes (1.57 GiB/s). The underlying disk can read more than 3 Gib/s, but even with
nbdcopy, using libnbd async API using queue depth of 8, it takes 2.4 seconds (2.5 GiB/s).

Detecting zeroes in client side can be effective if we upload via very slow network
(e.g. 1 Gbit), when upload throughput is ~120 MiB/s. In this case uploading without
zero detection will take 50 seconds, and with zero detection the operation will take
about 4 seconds, which is 12 times faster.

nirs (commented):
bench: Start benchmark directory should be posted separately. It can be merged now regardless of zero detection.

nirs:
Stale for long time, feel free to take over.

PR #42, opened 2022-03-13
qemu_nbd: Select cache and aio automatically

nirs:
Previously we use --cache=writeback and --aio=threads so we can access
images on file system that does not support direct I/O. This is
sometimes faster than --cache=none and --aio=native, but typically give
less consistent results, and pollute the page cache with image data,
which is mostly likely will never be used.

When uploading and downloading images on a hypervisor, using the page
cache can be harmful and cause sanlock timeouts when the kernel try to
flush huge images to the underlying storage.

Future qemu-img and qemu-nbd are expected to implement "auto" cache and
aio modes, selecting --cache=none and --aio=native if possible. This
change implements this in our qemu_nbd wrapper so we can use this now
with current qemu version.

When starting qemu_nbd.Server(), if cache was not specified, we select
cache="none" if the image can be opened with direct I/O. If aio was not
specified, we select it based on the cache value.

### DIFF
--- a/ovirt_imageio/_internal/qemu_nbd.py
+++ b/ovirt_imageio/_internal/qemu_nbd.py
@@ -28,7 +28,7 @@ class Server:
 
     def __init__(
             self, image, fmt, sock, export_name="", read_only=False, shared=8,
-            cache="none", aio="native", discard="unmap", detect_zeroes="unmap",
+            cache=None, aio=None, discard="unmap", detect_zeroes="unmap",
             bitmap=None, backing_chain=True, offset=None, size=None,
             timeout=10.0):
         """
@@ -41,8 +41,11 @@ class Server:
             export_name (str): expose export by name
             read_only (bool): export is read-only
             shared (int): export can be shared by specified number of clients
-            cache (str): cache mode (none, writeback, ...)
-            aio (str): AIO mode (native or threads)
+            cache (str): cache mode (none, writeback, ...). If not specified,
+                probe image and use "none" if direct I/O is supported, and
+                "writeback" if not.
+            aio (str): AIO mode (native or threads). If not specified, use
+                "native" if cache is "none" and "threads" otherwise.
             discard (str): discard mode (ignore, unmap)
             detect_zeroes (str): Control the automatic conversion of plain zero
                 writes by the OS to driver-specific optimized zero write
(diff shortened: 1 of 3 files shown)


nirs:
Tested using backup stress test:
https://gitlab.com/nirs/ovirt-stress/-/tree/master/backup

2022-03-13 10:37:49,122 INFO    (MainThread) 64 full backups, 896 incremental backups, 0 failed, 960 passed, 0 errored in 333.8 minutes

The test use imageio client to download backups and created checksum of the backup. With this
change all the backups use --cache=none --aio=native on the client side.

ovirt_imageio/_internal/qemu_nbd.py
```diff
@@ -41,8 +41,11 @@ def __init__(
             export_name (str): expose export by name
             read_only (bool): export is read-only
             shared (int): export can be shared by specified number of clients
-            cache (str): cache mode (none, writeback, ...)
-            aio (str): AIO mode (native or threads)
+            cache (str): cache mode (none, writeback, ...). If not specified,
+                probe image and use "none" if direct I/O is suported, and
```
bennyz, line 45:
```suggestion
                probe image and use "none" if direct I/O is supported, and
```

ovirt_imageio/_internal/qemu_nbd.py
```diff
@@ -86,6 +89,15 @@ def url(self):
         return urllib.parse.urlparse(url)
 
     def start(self):
+        # Implemnet "auto" cache mode, planned for qemu 7.
```
bennyz, line 92:
```suggestion
        # Implement "auto" cache mode, planned for qemu 7.
```

bennyz (dismissed):

bennyz (approved):

ovirt_imageio/_internal/qemu_nbd.py
```diff
@@ -86,6 +89,15 @@ def url(self):
         return urllib.parse.urlparse(url)
 
     def start(self):
+        # Implement "auto" cache mode, planned for qemu 7.
+        if self.cache is None:
+            self.cache = "none" if self._can_use_direct_io() else "writeback"
+            log.debug("Using cache=%r", self.cache)
```
barpavel, line 95: Shouldn't this `log.debug()` be out of the `if` (so it will be printed **always**)?
Same for line #99.

barpavel (commented):
All looks good, just 1 question about 2 logs.

ovirt_imageio/_internal/qemu_nbd.py
nirs: No, this is logged only if we cache was automatically set. If the user set the cache to explicit value, we use it as is so there is no need to log.

PR #50, opened 2022-03-20
http: Increase listen backlog to 40

nirs:
Both TCP and unix socket based servers inherited the listen backlog from
python socketserver.TCPServer. The default value (5) was never a
problem, but new automated tests using 16 concurrent connections started
to fail randomly recently with:

ERR: Traceback (most recent call last):
  ...
  File "ovirt_imageio/_internal/backends/http.py", line 392, in _create_unix_connection
    con.connect()
  File "ovirt_imageio/_internal/backends/http.py", line 627, in connect
    self.sock.connect(self.path)
BlockingIOError: [Errno 11] Resource temporarily unavailable

Increase the backlog to 40. We recommend to do up to 10 image transfers
per host. Doing more will work, but if someone will try to start 60
connections at the same time they will likely experience the same error,
which is good sign that they are overloading the system.

Bug-Url: https://bugzilla.redhat.com/2066113


ovirt_imageio/_internal/http.py
```diff
@@ -72,6 +72,11 @@ class Server(socketserver.ThreadingMixIn,
     """
     daemon_threads = True
 
+    # Override socketserver.TCPServer default (5) to allow larger number
+    # of concurrent connections. If larger number of connections are
+    # made, some connections will fail with EAGAIN.
+    request_queue_size = 40
```
vjuranek, line 78: Would be useful to make it configurable instead of hardcoding another value? If we want to stick with max 10 concurrent image download, it would be good to add a comment about it here.

bennyz (approved):

ovirt_imageio/_internal/http.py
nirs: Configuration is a problem since this server is not designed to support many concurrent users. The hard coded limit makes it hard to abuse it.

But a comment about supporting 10 concurrent transfers is good idea.

vjuranek (approved):

PR #59, opened 2022-04-01
docker: Add python 3.8 to el8 container

nirs:
We want to build also python 3.8 version of the common and client
packages, needed by ovirt python sdk. Adding python 3.8 to the container
enables the build and the tests.

python38-systemd is not available, hopefully this is a pure python
package and python3-systemd will works for the tests.

### DIFF
--- a/docker/Dockerfile.centos-8
+++ b/docker/Dockerfile.centos-8
@@ -18,10 +18,13 @@ RUN echo v1 \
         python3-devel \
         python3-pip \
         python3-systemd \
+        python38-devel \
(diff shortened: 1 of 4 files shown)


mwperina (dismissed):
+1

nirs:
This breaks the tests when using the new container:
https://github.com/nirs/ovirt-imageio/runs/5789894603?check_suite_focus=true

We have 2 issues:
- ioutil module not found when running the test-py38 env
- systemd is not available, since there is no python38-systemd module

PR #77, opened 2022-05-29
examples: Show how to use client.checksum()

nirs:
Add simple example showing how to use client.checksum() to compute a
checksum for local image.

This is mainly useful for benchmarking, for example comparing to
sha256sum, or blksum.

### DIFF
--- a/examples/checksum
+++ b/examples/checksum
@@ -1,0 +1,18 @@
+#!/usr/bin/python3
+# ovirt-imageio
+# Copyright (C) 2018 Red Hat, Inc.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+
+"""
+Show how to compute disk image checksum.
+"""
+import sys
+from ovirt_imageio import client
+
+path = sys.argv[1]
+res = client.checksum(path, algorithm="sha256")
+print(f"{res['checksum']}  {path}")


examples/checksum
```diff
@@ -0,0 +1,18 @@
...
+# (at your option) any later version.
+
+"""
+Show how to compute disk image checksum.
+"""
+import sys
+from ovirt_imageio import client
+
+path = sys.argv[1]
+res = client.checksum(path, algorithm="sha256")
```
aesteve-rh, line 17: `path` is not checked inside `checksum`, so maybe would be good to check if path exists before passing it.

Also, not in the scope of this PR, but the `algorithm` parameter here would benefit of it being an `Enum` instead of a string.

vjuranek: It's trivial example, so not sure if it worth to complicate it with checking the path. But yes, it's definitely good practice to validate any user input, so +1.

vjuranek: > The purpose of this is to be able to benchmark the imageio implementation with blksum

In such case I'd consider to create `benchmarks` dir or something like this to be clear that the programs there are used for benchmarking and therefore can skip some steps which may be included in "normal" script which uses imageio and is exposed to common users.

nirs: See the other examples (e.g. examples/imageio-client). Adding unneeded error checking does not help to make examples better.

nirs: > > The purpose of this is to be able to benchmark the imageio implementation with blksum
> 
> In such case I'd consider to create `benchmarks` dir or something like this to be clear that the programs there are used for benchmarking and therefore can skip some steps which may be included in "normal" script which uses imageio and is exposed to common users.

But this directory is called "examples", and are part of the source. They are not installed by any package so they are not exposed to common users.

aesteve-rh (approved):
Understood in the comments. All good.

bennyz (approved):

examples/checksum
vjuranek: > But this directory is called "examples", and are part of the source. They are not installed by any package so they are not exposed to common users.

Sure, but are example how to use imageio. If I see user input passing without any validation into imageio in the example, my expectation would be that the imegio client does the validation, which is not the case and it's passed to qemu nbd. In this case it cannot do any harm, but in general mixing benchmarking script and example scripts is not good as it can create false expectation or result into poor 3rh party programs using imageio because authors will use examples as a reference and will copy&paste them into their code.

I don't mean it in a sense that this particular example has to be fixed, but as an idea for the future that examples scripts and benchmark scripts are different in their purpose and IMHO shouldn't be fixed.

vjuranek (approved):

examples/checksum
nirs: The case of validating the path is more tricky - qemu-img and qemu-nbd (tools used
under the hood) support a url (e.g. http://...) or json description of the image. With the current code we support only filename, but we could support other formats. So even in a real command line tool I'm not sure the tool should do any validation. The validation is running "qemu-img info" with the path.

I think this discussion show that we need to improve how imageio reports errors when running qemu-img, and we are missing a README for this directory.

nirs: Filed #79 and #78 so we don't forget about this.

PR #82, opened 2022-06-12
Containerized deployment

nirs:
Initial work for deploying ovirt-imageio as containerized workload.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ GENERATED = \
 
 METADATA = ovirt_imageio/_internal/version.py Makefile
 
-.PHONY: build check dist srpm rpm clean images clean-images storage clean-storage $(SPEC_NAME) venv
+.PHONY: build check dist container srpm rpm clean images clean-images storage clean-storage $(SPEC_NAME) venv
 
 build:
 	python3 setup.py build_ext --build-lib .
@@ -24,6 +24,11 @@ check: images
 
 dist: $(GENERATED)
 	python3 setup.py sdist --dist-dir "$(OUTDIR)"
+
+container: dist
+	cp $(OUTDIR)/ovirt-imageio-*.tar.gz container/ovirt-imageio.tar.gz
+	podman build -t ovirt-imageio container
+	rm -f container/ovirt-imageio.tar.gz
 
 srpm: dist
 	rpmbuild --define="_topdir $(RPM_TOPDIR)" \
--- a/Makefile
+++ b/Makefile
@@ -27,8 +27,9 @@ dist: $(GENERATED)
 
 container: dist
 	cp $(OUTDIR)/ovirt-imageio-*.tar.gz container/ovirt-imageio.tar.gz
+	virt-builder fedora-36 --output container/disk.img --format qcow2
 	podman build -t ovirt-imageio container
-	rm -f container/ovirt-imageio.tar.gz
+	rm -f container/ovirt-imageio.tar.gz container/disk.img
 
 srpm: dist
 	rpmbuild --define="_topdir $(RPM_TOPDIR)" \
--- a/container/Containerfile
+++ b/container/Containerfile
@@ -1,0 +1,20 @@
+# Build stage
+
+FROM alpine AS build
+RUN apk add --no-cache --update build-base linux-headers python3 python3-dev py3-pip
+WORKDIR /build
+COPY ovirt-imageio.tar.gz .
+RUN python3 -m venv /venv
+RUN /venv/bin/pip install ovirt-imageio.tar.gz
+
+# Final stage
+
+FROM alpine
+RUN apk add python3
+RUN addgroup --system ovirtimg
+RUN adduser --system --ingroup ovirtimg ovirtimg
+RUN mkdir /run/ovirt-imageio /etc/ovirt-imageio
+COPY --from=build /venv /venv
+COPY conf.d /etc/ovirt-imageio/conf.d
+EXPOSE 80/tcp
+CMD /venv/bin/ovirt-imageio
--- a/container/Containerfile
+++ b/container/Containerfile
@@ -19,5 +19,6 @@ COPY conf.d /etc/ovirt-imageio/conf.d
 COPY file.json file.json
 COPY disk.img /images
 RUN chown ovirtimg:ovirtimg /images/disk.img
+VOLUME [ "/pki/system" ]
 EXPOSE 80/tcp
 CMD /venv/bin/ovirt-imageio -t file.json
--- a/container/Containerfile
+++ b/container/Containerfile
@@ -13,8 +13,11 @@ FROM alpine
 RUN apk add python3
 RUN addgroup --system ovirtimg
 RUN adduser --system --ingroup ovirtimg ovirtimg
-RUN mkdir /run/ovirt-imageio /etc/ovirt-imageio
+RUN mkdir /run/ovirt-imageio /etc/ovirt-imageio /images
 COPY --from=build /venv /venv
 COPY conf.d /etc/ovirt-imageio/conf.d
+COPY file.json file.json
+COPY disk.img /images
+RUN chown ovirtimg:ovirtimg /images/disk.img
 EXPOSE 80/tcp
-CMD /venv/bin/ovirt-imageio
+CMD /venv/bin/ovirt-imageio -t file.json
--- a/container/conf.d/daemon.conf
+++ b/container/conf.d/daemon.conf
@@ -1,0 +1,18 @@
+[daemon]
+systemd_enable = false
+
+[tls]
+enable = false
+
+[remote]
+port = 80
+
+[local]
+enable = false
+
+[handlers]
+keys = stderr
+
+[logger_root]
+level = INFO
+handlers = stderr
--- a/container/conf.d/daemon.conf
+++ b/container/conf.d/daemon.conf
@@ -2,7 +2,9 @@
 systemd_enable = false
 
 [tls]
-enable = false
+key_file = /pki/system/key.pem
+cert_file = /pki/system/cert.pem
+ca_file = /pki/system/ca.pem
 
 [remote]
 port = 80
--- a/container/file.json
+++ b/container/file.json
@@ -1,0 +1,8 @@
+{
+    "uuid": "ticket-id",
+    "size": 6442450944,
+    "url": "file:///images/disk.img",
+    "timeout": 3000,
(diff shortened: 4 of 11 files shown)


vjuranek (dismissed):
This looks like a very good idea. One question: how do you plan to do concurrent downloads? Local port `8000` will be bound to already running container, so you will have to use different port. Or maybe the question should be: what is the final goal - use imageio in env. with orchestrations like in k8s/openshit, make to easy tests for developers, something else?

nirs:
> This looks like a very good idea. One question: how do you plan to do concurrent downloads? 
Local port `8000` will be bound to already running container, so you will have to use different port. Or maybe the question should be: what is the final goal - use imageio in env. with orchestrations like in k8s/openshit, make to easy tests for developers, something else?

Port 8000 is not very important, actually it should be port 80 since we start as root
in the container, and drop privileges.

When running in a container, the container runtime will publish the port to whatever
you want. The container will be used for one transfer and then it will be discarded.

To transfer many images at the same time the container runtime will start as many 
instances as needed and assign some port that will be used by the client. If we think
about Kubernetes, every transfer pod will be exposed outside using a service.

vjuranek:
> To transfer many images at the same time the container runtime will start as many instances as needed and assign some port that will be used by the client. If we think about Kubernetes, every transfer pod will be exposed outside using a service.

Basically the question was how the client will learn about imageio port. Is it intended for k8s where we have service for it or something else? I know implementation of any such orchestration (if not run under k8s) is outside the scope of this PR, just wondering what is the motivation for this.

nirs:
> > To transfer many images at the same time the container runtime will start as many instances as needed and assign some port that will be used by the client. If we think about Kubernetes, every transfer pod will be exposed outside using a service.
> 
> Basically the question was how the client will learn about imageio port. Is it intended for k8s where we have service for it or something else? I know implementation of any such orchestration (if not run under k8s) is outside the scope of this PR, just wondering what is the motivation for this.

In k8s the port can be port 80 (as it should really be) since every pod has its own ip address. The client can learn about the port and ip address in the same way it works
in oVirt - you start a transfer and the system provides a url where you can upload
or download. The fact that in oVirt we use the same port is implementation detail,
not part of the API.

We could also use the same model in oVirt, in the same way we start new qemu-nbd for
every transfer we could start a new instance of (a simpler) ovirt-imageio-daemon that
connects to single image or unix socket. This can improve performance and provide
isolation so stuck storage can harm only one instance.

container/conf.d/daemon.conf
```diff
@@ -0,0 +1,18 @@
+[daemon]
+systemd_enable = false
+
+[tls]
+enable = false
+
+[remote]
+port = 8000
```
nirs, line 8: Change to 80 - our protocol is http so we should listen to the standard ports.

container/conf.d/daemon.conf
```diff
@@ -0,0 +1,18 @@
+[daemon]
+systemd_enable = false
+
+[tls]
+enable = false
```
nirs, line 5: This is temporary until we have configure PKI via files mounted files in the container, but it can also be used if the server is behind a reverse proxy (e.g. k8s ingress) that will use https and communicate via http with the node.

aesteve-rh:
For some reason I missed the notification for this branch. I'll check it in deep next week!

aesteve-rh (dismissed):

container/Containerfile
```diff
@@ -13,8 +13,11 @@ FROM alpine
 RUN apk add python3
 RUN addgroup --system ovirtimg
 RUN adduser --system --ingroup ovirtimg ovirtimg
-RUN mkdir /run/ovirt-imageio /etc/ovirt-imageio
+RUN mkdir /run/ovirt-imageio /etc/ovirt-imageio /images
 COPY --from=build /venv /venv
 COPY conf.d /etc/ovirt-imageio/conf.d
+COPY file.json file.json
+COPY disk.img /images
```
aesteve-rh, line 20: I will change it to `VOLUME` or `CMD --mount` later. I thought it would be nice to copy the image to have a ticket with well-known size and name. But the ticket size can be modified before invoking `make container`, and mounting the image with `VOLUME` allows for a more portable (smaller) container image.

nirs: The ticket and the image should not be part of the container.

In k8s the ticket will be provided by configmap, same way engine details will be provided. The disk will be provided via a pvc.

Makefile
```diff
@@ -27,8 +27,9 @@ dist: $(GENERATED)
 
 container: dist
 	cp $(OUTDIR)/ovirt-imageio-*.tar.gz container/ovirt-imageio.tar.gz
+	virt-builder fedora-36 --output container/disk.img --format qcow2
```
nirs, line 30: Including the disk image in the container is really bad. It should be provided via --volume in the command line by the user of the container:

    podman run --volume /path/to/disk.img:/disk.img:Z ...
    
Creating the disk image each time we create a container is bad. The image should be created by the user of the container (e.g. tests) once before running the tests.

And remember that we have 2 use cases:
- download existing disk - can be a file or block device on the host
- upload to empty disk - can be a file or block device on the host

container/Containerfile
```diff
@@ -13,8 +13,11 @@ FROM alpine
...
+RUN chown ovirtimg:ovirtimg /images/disk.img
```
nirs, line 21: This assumes that the images is always called disk.img, which is not a good assumption. The file name and location inside the container should be dynamic. The ticket will point to the location of the file inside the container.

Onwership and permissions should be set by the user of the container so imageio can access the image. This should not be part of the container file.

container/Containerfile
```diff
@@ -13,8 +13,11 @@ FROM alpine
...
 EXPOSE 80/tcp
-CMD /venv/bin/ovirt-imageio
+CMD /venv/bin/ovirt-imageio -t file.json
```
nirs, line 24: The name of the ticket and its location should be set by the user of the container. One way to pass the info is an environment variable.

container/file.json
```diff
@@ -0,0 +1,8 @@
+{
+    "uuid": "ticket-id",
+    "size": 6442450944,
+    "url": "file:///images/disk.img",
+    "timeout": 3000,
+    "sparse": true,
+    "ops": ["read", "write"]
+}
```
nirs, line 8: This ticket can be part of the tests.

container/Containerfile
```diff
@@ -19,5 +19,6 @@ COPY conf.d /etc/ovirt-imageio/conf.d
 COPY file.json file.json
 COPY disk.img /images
 RUN chown ovirtimg:ovirtimg /images/disk.img
+VOLUME [ "/pki/system" ]
```
nirs, line 22: Where this directory comes from?

I think we can ignore tls for now, better leave it to next step. It will be good enough to have a working container supporting only http as first step.

In k8s the container will likely use http, and https will be handled by ingress (e.g. nginx).

nirs:
Please post a new pr instead of modifying this pr. I cannot request changes in this pr :-)

container/Containerfile
aesteve-rh: If I understood correctly the docs, this creates a mount point at `/pki/system`, so that it can be mounted at runtime by doing `podman run -v /source/folder:/pki/system`. Alternatively, `CMD --mount` (bind), would require setting a source folder at build time, which is more inconvenient, unless we obtain it with an env variable (maybe).

So I opted for the `VOLUME` solution. But I am exploring possibilities, and any feedback is welcome.

aesteve-rh:
I'll close the PR then, and post a new one.

container/Containerfile
aesteve-rh: The location could be pre-defined and let the user pick only the name through an env variable.

So we have:
```Containerfile
ARG TICKET_NAME # or ARG TICKET_NAME=ticket.json (for default value)
VOLUME [ "ticket" ]
CMD /venv/bin/ovirt-imageio -t /ticket/${TICKET_NAME}
```

And it can be run by doing:
```
TICKET_NAME=ticket.json podman run \
    --env-host \
    --rm \
    --privileged \
    -it -p 8080:80 \
    -v ./path/to/ticket:/ticket
    -v ./path/to/images/:/path/of/the/ticket/img \
    localhost/ovirt-imageio:latest
```

Otherwise, we have to set the whole path of the file in an env variable, or have two, one for the folder and one for the name. And we only need one ticket every time it is run, so I find it easier to just assume a folder that will contain the ticket, and let the user specify which ticket file to load.

There is not much difference, I tried both, and it works either way. Just feels easier for users if we assume a mount point in the container for tickets. Also, I don't think it is interesting to try different locations for tickets (it may be more interesting for images), so shouldn't be a problem to have it pre-defined.

nirs: I'm not sure why we need this:

    -v ./path/to/images/:/path/of/the/ticket/img

Instead of

    -v /path/to/image:/disk:Z

But it is true that using well known path for the ticket makes it easier to use. For the image I'm not sure we can require certain path since in kubvirt they always use /disk.img and we want to be compatible with this. On the other hand we don't want to require this location.

For k8s the ticket is likely to be passed using configmap, and in this case the configmap keys are converted to mounts inside the container, but I'm not sure how it looks inside the container.

In summary, we want to have a flexible solution that can work for any use case. Making it easier to use for some cases is an optimization that can be done later.

container/Containerfile
nirs: Not sure what VOLUME gives us - if you use:

    podman run -v /source/folder:/pki/system

It works without specifying VOLUME in the container file.

container/Containerfile
aesteve-rh: Just for the record, I was just talking on defining the ticket path, the image path I agree that should remain flexible, and can be determined by the information in the ticket and the `--volume` specified at runtime.

This was just an example:
```
-v ./path/to/images/:/path/of/the/ticket/img
```
I wanted to put the focus on the ticket volume. Your counterexample looks good to me too.

container/Containerfile
aesteve-rh: I was not sure, I thought it was just to set an expected mount point inside the container. Actually, declaring a VOLUME in a containerfile creates an anonymous persistent volume that can be found with `podman volume ls` if you do not specify a runtime volume. This can be later shared between containers if you use `--volume-from`.

I think it may not be interesting for our usecase. In fact, I have found this article that discourages the use of volumes in dockerfiles: https://boxboat.com/2017/01/23/volumes-and-dockerfiles-dont-mix/

PR #85, opened 2022-06-16
version: Bump to 2.4.6

nirs:
Start next development cycle.

### DIFF
--- a/ovirt_imageio/_internal/version.py
+++ b/ovirt_imageio/_internal/version.py
@@ -6,7 +6,7 @@
 # the Free Software Foundation; either version 2 of the License, or
 # (at your option) any later version.
 
-string = "2.4.5"
+string = "2.4.6"
 
 if __name__ == "__main__":
     print(string)


aesteve-rh (approved):
It is a bit odd that the variable containing the version number is called `string`. It is not a reserved word, nor a built-in type, so it should be all good. But it is not the most informative variable name.

But is just a note. Changing this variable name is not in the scope of this PR, so I'll approve.

nirs:
> It is a bit odd that the variable containing the version number is called `string`. It is not a reserved word, nor a built-in type, so it should be all good. But it is not the most informative variable name.

ovirt_imageio version string is kept in:

    ovirt_imageio.version.string

Isn’t the best name possible?

python has namespaces - when you think about names you should always consider the full name.

aesteve-rh:
With the complete name it looks better, true. Is one of the sides of Python that still gets me sometimes...

Still, I would probably go for the semi-standard name `__version__`, but is just a matter of taste.

nirs:
> Still, I would probably go for the semi-standard name `__version__`, but is just a matter of taste.

`__version__` is useful when you have a single module. In a package we could add it to `__init__.py`, but then it would be harder to import the version in modules using the version.

We can expose the version as `__version__` in `__init__py` to make it easy to query the package version.

PR #87, opened 2022-06-26
Build cleanups

nirs:
- Fix error handling in build-aux/release script
- Remove unneeded workaround for copr build

Example copr build with from this change:
https://copr.fedorainfracloud.org/coprs/nsoffer/ovirt-imageio-preview/build/4565080/

### DIFF
--- a/build-aux/release
+++ b/build-aux/release
@@ -1,18 +1,28 @@
-#!/bin/sh
+#!/usr/bin/python3
 
-# Is this a release - build from tag?
-if git describe --tags --match "v[0-9]*" | grep -q '-'; then
-    # This is a master build, we want to make every build
-    # newer than all the previous builds using a timestamp,
-    # and make it easy to locate the commit from the build
-    # with the git commit hash.
-    release="0.$(date -u +%Y%m%d%H%M)"
-    commithash=$(git rev-parse --short HEAD)
-    release+=.git$commithash
-else
-    # We want to make tagged build newer than any other build from same
-    # version-release.
-    release="1"
-fi
+import datetime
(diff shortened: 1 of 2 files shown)


nirs:
A copr build from this PR:
https://copr.fedorainfracloud.org/coprs/nsoffer/ovirt-imageio-preview/build/4565094/

aesteve-rh (dismissed):
Using `GitPython` would provide nice methods to get the `hexsha` of a git object, and many other commands such as `describe`, without having to manually hack getting clean data with `subprocess`, the `decode` and the `strip`. 

Furthermore, since you are basically only interested in the `stdout`, even if you stick with `subprocess`, I would use [check_output](https://docs.python.org/3/library/subprocess.html#subprocess.check_output) instead of `run`. In any case, no need to pipe also the stderr, as it just gets ignored.

But the script is small enough so that it won't matter that much, so I'll approve, and you decide if you want to change it or not.

vjuranek (approved):

nirs:
copr build from this pr:
https://copr.fedorainfracloud.org/coprs/nsoffer/ovirt-imageio-preview/build/4566209/

build-aux/release
```diff
@@ -1,18 +1,28 @@
...
-    # version-release.
-    release="1"
-fi
+import datetime
+import subprocess
 
-echo $release
+
+def run(*args):
+    return subprocess.check_output(args).decode().strip()
```
aesteve-rh, line 8: Nice little change with the variable `*args`!

aesteve-rh (approved):

Issue #31, opened 2022-02-06
Add unit constants


nirs:
Instead of repeating 1024**2, 1024**3, ... use readable MiB, GiB constants.

We can borrow this file from vdsm (borrowed from sanlock):
https://github.com/oVirt/vdsm/blob/master/lib/vdsm/common/units.py

PR #88, opened 2022-07-05
imageio: add unit constants

aesteve-rh:
There are many repetitions of `1024**2` and
`1024**3` for MiB and GiB, respectively.

Instead, add constants and use them
to make it more readable.

Fixes: #31


nirs (approved):
Thanks!

PR #92, opened 2022-07-11
Always build ioutil for platform python

nirs:
When running py38 tests, we run ovirt-imageio using /usr/bin/python3,
but we built ioutil C extension with "python", which is python 3.8. We
cannot fix this by running the ovirt-imageio with python 3.8 since we
don't support python 3.8 for for the server.

Change tox to build the C extension for both current python version and
platform python.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -20,7 +20,9 @@ deps =
     test: pytest-cov
     test: pytest-timeout
 commands_pre =
+    # Build ioutil for both current python (e.g. 3.8) and platform python (e.g. 3.6)
     python setup.py build_ext --build-lib .
+    /usr/bin/python3 setup.py build_ext --build-lib .
     python -c 'from test import testutil; print("ipv6 supported: %s" % testutil.ipv6_enabled())'
 commands =
     test: pytest -m 'not benchmark' --cov=ovirt_imageio --durations=10 {posargs}


tox.ini
```diff
@@ -20,7 +20,9 @@ deps =
     test: pytest-cov
     test: pytest-timeout
 commands_pre =
+    # Build ioutil for both current python (e.g. 3.8) and platform python (e.g. 3.6)
     python setup.py build_ext --build-lib .
+    /usr/bin/python3 setup.py build_ext --build-lib .
```
barpavel, line 25: Will the 2 `setup.py build_ext --build-lib .` have different names?

nirs: yes:

```
gcc -pthread -shared -Wl,-z,relro -Wl,-z,now -g -Wl,-z,relro -Wl,-z,now -g build/temp.linux-x86_64-3.8/ovirt_imageio/_internal/ioutil.o -L/usr/lib64 -o ./ovirt_imageio/_internal/ioutil.cpython-38-x86_64-linux-gnu.so
gcc -pthread -shared -Wl,-z,relro -Wl,-z,now -g -Wl,-z,relro -Wl,-z,now -g build/temp.linux-x86_64-3.6/ovirt_imageio/_internal/ioutil.o -L/usr/lib64 -lpython3.6m -o ./ovirt_imageio/_internal/ioutil.cpython-36m-x86_64-linux-gnu.so
```

barpavel (approved):

tox.ini
barpavel: One more "corner case" :)
What if current Python version and platform Python versions are the same - won't there be a problem with `ioutil.cpython` override when running the newly added command?

nirs: No, build_ext does not nothing in this case.

See the latest run on centos 8:
https://github.com/oVirt/ovirt-imageio/runs/7287479259?check_suite_focus=true

vjuranek (approved):

aesteve-rh (approved):

Issue #12, opened 2021-12-13
Simplify ipv6 setup in tests


nirs:
Currently we use configure docker and restart the service, and then
run the tests using docker:

    - name: Configure docker
      run: sudo ./ci/configure-docker.py
    - name: Pull container
      run: docker pull quay.io/ovirt/imageio-test-${{matrix.distro}}
    - name: Run tests
      run: |
        docker run \
        --privileged \
        --rm \
        --volume `pwd`:/src:Z \
        quay.io/ovirt/imageio-test-${{matrix.distro}} \
        bash -c "cd /src && ci/test.sh"

We can simplify this to use a privilege container and enable ipv6 from the container
like:
https://github.com/oVirt/vdsm/blob/master/automation/tests.sh

https://github.com/oVirt/vdsm/commit/f937bc31f686698c208b04e26a989ff8d821ba5f

PR #93, opened 2022-07-13
ci: simplify ipv6 setup

aesteve-rh:
IPv6 is disabled in docker containers on GH actions.
Enable IPv6 and simplify the ci workflow job for
running tests so that we do not need to configure
docker beforehand.

Fixes: #12


.github/workflows/ci.yml
```diff
@@ -15,20 +15,14 @@ jobs:
         - centos-9
         - fedora-35
         - fedora-36
+    container:
+      image: quay.io/ovirt/imageio-test-${{matrix.distro}}
+      # Required to create loop devices.
+      options: --privileged
```
nirs, line 21: Nice, I think this was not possible when we add the CI, or maybe we missed this option.
vdsm ci can be also be simplified using this.

.github/workflows/ci.yml
```diff
@@ -15,20 +15,14 @@ jobs:
...
     steps:
     - uses: actions/checkout@v2
-    - name: Configure docker
-      run: sudo ./ci/configure-docker.py
```
nirs, line 21: Would be nice to delete the helper script in the next PR.

nirs (approved):

Issue #90, opened 2022-07-09
Separate ci job for centos-8 python 3.8


nirs:
On centos 8 we run the tests twice both with python 3.6 (platform python) and python 3.8. This
doubles the run time  and make it harder to check failures.

Create separate builds for centos 8:
- tox -e flake8,test-py36,bench-py36
- tox -e flake8,test-py38,bench-py38

This will also avoid the annoying SKIPPED lines in the summary:

```
___________________________________ summary ____________________________________
  flake8: commands succeeded
SKIPPED:  test-py36: InterpreterNotFound: python3.6
SKIPPED:  test-py37: InterpreterNotFound: python3.7
SKIPPED:  test-py38: InterpreterNotFound: python3.8
  test-py39: commands succeeded
SKIPPED:  test-py310: InterpreterNotFound: python3.10
SKIPPED:  bench-py36: InterpreterNotFound: python3.6
SKIPPED:  bench-py37: InterpreterNotFound: python3.7
SKIPPED:  bench-py38: InterpreterNotFound: python3.8
  bench-py39: commands succeeded
SKIPPED:  bench-py310: InterpreterNotFound: python3.10
  congratulations :)
```

It will be:
```
  flake8: commands succeeded
  test-py39: commands succeeded
  bench-py39: commands succeeded
  congratulations :)
```

PR #99, opened 2022-07-15
ci: split centos 8 jobs for py36 and py38

aesteve-rh:
On centos 8 we run the tests twice both with python 3.6
(platform python) and python 3.8. This doubles the
run time and makes it harder to check failures.
Other distros only run a single python environment
tests, but checks for all of them, creating a lot
of unnecesary noise in the logs.

Separate centos-8 enviornments for py36 and py38
to allow parallel execution. Other distributions
should run only a specific python environment
that matches the one they have installed.

Fixes: https://github.com/oVirt/ovirt-imageio/issues/90

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,11 +10,17 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        distro:
-        - centos-8
-        - centos-9
-        - fedora-35
-        - fedora-36
+        include:
+          - distro: centos-8
+            env: py36
+          - distro: centos-8
+            env: py38
+          - distro: centos-9
+            env: py39
+          - distro: fedora-35
+            env: py310
+          - distro: fedora-36
+            env: py310
     container:
       image: quay.io/ovirt/imageio-test-${{matrix.distro}}
       # Required to create loop devices.
@@ -22,7 +28,7 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: Run tests
-      run: ci/test.sh
+      run: ci/test.sh ${{matrix.env}}
   rpm:
     runs-on: ubuntu-latest
     strategy:
--- a/ci/test.sh
+++ b/ci/test.sh
@@ -21,5 +21,13 @@ git config --global --add safe.directory "$(pwd)"
 # Enable IPv6
 echo 0 > /proc/sys/net/ipv6/conf/all/disable_ipv6
 
+if [ -z $1 ]; then
+    echo "Missing argument 'ENV' (tox environment)."
+    echo "Usage: $0 ENV"
+    exit 1
+fi
+
+env="$1"
+
 make storage
-make check
+tox -e flake8,test-$env,bench-$env


aesteve-rh:
For this split, we will need to change project setting to remove centos-8 from the required jobs, and add one or both the environment specific centos-8 jobs instead.

.github/workflows/ci.yml
```diff
@@ -11,18 +11,22 @@ jobs:
...
         distro:
-        - centos-8
         - centos-9
         - fedora-35
         - fedora-36
+        include:
+          - distro: centos-8
+            env: py36
+          - distro: centos-8
+            env: py38
```
nirs, line 21: I hope we can make this simpler - list of distro/env combinations:

    cs8-py36:
        disto: centos-8
        env: py36
    cs9-py38:
        disto: centos-8
        env: py38
    cs9-py39
        disto: centos-9
        env: py39
    f35-py310:
        disto: fedora-35
        env: py310
    f36-py310:
        disto: fedora-36
        env: py310

.github/workflows/ci.yml
```diff
@@ -11,18 +11,22 @@ jobs:
...
     container:
       image: quay.io/ovirt/imageio-test-${{matrix.distro}}
       # Required to create loop devices.
       options: --privileged
     steps:
     - uses: actions/checkout@v2
     - name: Run tests
-      run: ci/test.sh
+      run: ci/test.sh ${{matrix.env}}
```
nirs, line 29: So matix.env si empty string when not specified?

This means we need to handle missing env in the script which will make it more complicated.

ci/test.sh
```diff
@@ -1,5 +1,25 @@
 #!/bin/bash -e
 
+ENVS="py36 py37 py38 py39 py310"
+
+print_help() {
+    echo "Usage: $0 [ENV]"
+    echo ""
+    echo "  Helper script to setup and run tests (all or ENV)."
```
nirs, line 8: We don't to run all envs in the ci, always one env. Runing all envs and skippnig missing python versions is convenience for developers so they can run "tox". The CI need more strict approach - it must specify a one env and the env must succeed.

ci/test.sh
```diff
@@ -21,5 +41,11 @@ git config --global --add safe.directory "$(pwd)"
...
 
+env=""
+if [ ! -z $1 ]; then
+    check_valid_env $1
+    env="-$1"
+fi
+
 make storage
-make check
+make check${env}
```
nirs, line 51: Run the test directly with tox instead:

    tox -e flake8,test-$env,bench-$env

Or change the makefile so we can pass the env:

    make check ENV=py38

If not specified, the makefile will run:

    tox

If specified, the makefile will run:

    tox -e flake8,test-$ENV,bench-$ENV

nirs (changes requested):

ci/test.sh
aesteve-rh: Using the envs for the makefile feels like an abuse. I prefer to run `tox` directly.

nirs: :+1:

.github/workflows/ci.yml
nirs: Lets go with include per distro.

aesteve-rh: Tried this in isolation with the last push, none of the other comments is addressed, but I wanted to see if it executed what we expected (and it does!). I'll cover the rest next week.

.github/workflows/ci.yml
```diff
@@ -10,19 +10,25 @@ jobs:
...
+          - distro: centos-8
+            env: py36
+          - distro: centos-8
+            env: py38
+          - distro: centos-9
+            env: py39
+          - distro: fedora-35
+            env: py310
+          - distro: fedora-36
+            env: py310
```
nirs, line 23: Nice! and it run correctly.

I wonder if include can be removed:

    matrix:
      - distro: ...
        env: ...
      ...

nirs (approved):

nirs:
@ahadas can you update the project settings required jobs to require test(distro, pyxx) instead of test(distro)?

We cannot merge this change without updating the project settings.

PR #100, opened 2022-07-18
Start the ovit-img tool.

nirs:
This is a minimal version supporting download disk. It works but it needs more work on user experience, logging, configuration, tuning options, and packaging.

To create the tool, the Go partial implementation was moved under the `go` directory, to
make room for the ovirt-img script at the top level.


ovirt_imageio/tool/__init__.py
```diff
@@ -0,0 +1,101 @@
...
+
+    cmd.add_argument(
+        "--cafile",
+        help="Path to ovirt-engine CA certificate")
+
+
+def fill_password(args):
+    if args.password_file:
+        with open(args.password_file) as f:
+            args.password = f.read().rstrip("\n")
```
aesteve-rh, line 79: Password as plain text is good enough here?

ovirt_imageio/tool/ovirt.py
```diff
@@ -0,0 +1,355 @@
...
+    log.debug("Using host id %s", host.id)
+
+    return host
+
+
+def create_transfer(
+        con, disk=None, direction=types.ImageTransferDirection.UPLOAD,
+        host=None, backup=None, inactivity_timeout=None, timeout=60,
+        disk_snapshot=None, shallow=None,
+        timeout_policy=types.ImageTransferTimeoutPolicy.CANCEL):
```
aesteve-rh, line 137: This is ok-ish, but when I see so many arguments in a function, where some of them are mutually exclusive (e.g., `disk` and `disk_snapshot`), I feel this could be a good place for data classes with inheritance.

This way we would receive only a single argument, the base class (polymorphism). And we will avoid having most of the if elses inside the function, as the data class will give us the correct ID, transfer, etc.

aesteve-rh (commented):
Looks good in general! See small comments below.

ovirt_imageio/tool/__init__.py
nirs: How your private key is protected? This password file can be protected in the way.

If users do not want to keep the password in a file, they can avoid --password-file. In this case the tool ask the password on every invocation using getpass().

An additional way we be using the keyring - but it is harder to implement and get work on multiple desktop environment.

When running in a k8s, the password will be provided as a secret, available only inside a pod.

aesteve-rh: If we don't want to use a hash algorithm to store the password (which may require to depend on a non-standard library, and more work than is worth it), we could also read environment variables where the password is stored.

That is a good mid-way strategy between prompting the user, which may be annoying specially for consecutive calls, and having the raw password stored in a file.

ovirt_imageio/tool/ovirt.py
nirs: Only disk and disk_snapshot are  mutually exclusive, and this is unfortunately the underlying engine API we have to work with.

I don't see how data classes can help, and they are not available in python 3.6 so
we cannot use them anyway.

In the original code when we added support for disk snapshots we did this in 
a backward compatible way so we don't break users using the existing helper
scripts, but we don't have this issue now since nobody is depending on this
and this is not a library but internal module.

How about using:

    create_tranfer(con, source=disk)
    create_tranfer(con, source=disk_snapshot)

And internally handle the type of the source?

    if isinstance(source, types.DiskSnapshot):
        transfer.snapshot = source
    elif isinstance(source, types.Disk):
        transfer.disk = source
    else:
        raise ValueError(...)

aesteve-rh: I mentioned dataclasses, but it can be solved with plain classes too (main difference is the explicit constructor anyway).

But nevermind that, what you proposed looks good too :)

ovirt_imageio/tool/__init__.py
nirs: We cannot hash the password, we need the plaintext for authentication. We can enctypt the password, but then you need the key to decrypt it.

ovirt_imageio/tool/__init__.py
```diff
@@ -0,0 +1,101 @@
...
```
rokkbert, line 78: Is that common practice in python, to [over]write cli-arguments like that? I think in some languages that is discouraged.

ovirt_imageio/tool/ovirt.py
```diff
@@ -0,0 +1,378 @@
...
+
+
+def find_storage_domain(con, disk):
+    service = _storage_domain_service(con, disk.storage_domains[0].id)
+    return service.get()
+
+
+def find_host(con, sd_name):
+    """
+    Check if we can preform a transfer using the local host and return a
```
rokkbert, line 70: s/preform/perform/

ovirt_imageio/tool/ovirt.py
```diff
@@ -0,0 +1,378 @@
...
+    host instance. Return None if we cannot use this host.
+
+    Using the local host for an image transfer allows optimizing the
+    connection using unix socket. This speeds up the transfer
+    significantly and minimizes the network bandwidth.
+
+    However using the local host is possible only if:
+    - The local host is a oVirt host
```
rokkbert, line 78: s/a oVirt/an oVirt/

ovirt_imageio/tool/__init__.py
nirs: args.password is not really an argument that a user can set. This is just an attribute in a Namespace object. But next version does not override it since we read the password just before it is needed.

ovirt_imageio/tool/options.py
```diff
@@ -0,0 +1,50 @@
...
+            "--username",
+            help="ovirt-engine username.")
+
+        cmd.add_argument(
+            "--password-file",
+            help="Read ovirt-engine password from file.")
+
+        cmd.add_argument(
+            "--cafile",
+            help="Path to ovirt-engine CA certificate")
```
aesteve-rh, line 43: This is a really nice argparse pattern you made here.

But, are these arguments added for each sub command? Can't we have them as global options already in the constructor, so we don't have to add them every time?

nirs: If we put the arguments on the parser they will not be reported as command options.

See how current backup_vm.py works:

```
$ ./backup_vm.py -h
usage: backup_vm.py [-h] -c CONFIG [--debug] [--logfile LOGFILE] {full,incremental,start,download,stop} ...

Backup VM disks

options:
  -c CONFIG, --config CONFIG
                        Use engine connection details from [CONFIG] section in ~/.config/ovirt.conf.
  --debug               Log debug level messages to logfile.
  --logfile LOGFILE     Log file name (default example.log).

commands:
  {full,incremental,start,download,stop}
    full                Run full backup.
    incremental         Run incremental backup.
    start               Start VM backup.
    download            Download VM backup disk.
    stop                Stop VM backup.
```

```
$ ./backup_vm.py full -h
usage: backup_vm.py full [-h] [--max-workers MAX_WORKERS] [--buffer-size BUFFER_SIZE] [--backup-dir BACKUP_DIR] [--timeout-policy {legacy,pause,cancel}] [--backup-uuid BACKUP_UUID]
                         [--disk-uuid DISK_UUID] [--description DESCRIPTION] [--skip-download]
                         vm_uuid

positional arguments:
  vm_uuid               UUID of the VM to backup.

options:
  --max-workers MAX_WORKERS
                        maximum number of workers to use for backup. The default (4) improves performance when backing up a single disk. You may want to use lower number if you back up
                        many disks in the same time.
  --buffer-size BUFFER_SIZE
                        Buffer size per worker. The default (4194304) gives good performance with the default number of workers. If you use smaller number of workers you may want use
                        larger value.
  --backup-dir BACKUP_DIR
                        Path to a directory to download backup disks to (The default is the current directory).
  --timeout-policy {legacy,pause,cancel}
                        The action to be made for a timed out transfer.
  --backup-uuid BACKUP_UUID
                        UUID of the created VM backup.
  --disk-uuid DISK_UUID
                        Disk UUID to backup. May be used multiple times to backup multiple disks. If not specified, backup all VM disks.
  --description DESCRIPTION
                        A description for the created backup/checkpoint to persist in the Engine DB.
  --skip-download       If specified, start and stop a backup without downloading the disks.
```

You cannot mix gloabl options in the command:

```
$ ./backup_vm.py full -c engine
usage: backup_vm.py [-h] -c CONFIG [--debug] [--logfile LOGFILE] {full,incremental,start,download,stop} ...
backup_vm.py: error: the following arguments are required: -c/--config
```

And you cannot see which options are available for single command.

When we add *all* the options only the sub parser it is much easier to use.

aesteve-rh: Oh snap. Ok thanks for showing.

[Click](https://click.palletsprojects.com/en/8.1.x/) handles this a bit more nicely. Just as a footnote, I know we don’t want to add this dependency, but is a good alternative that I used in the past.

ovirt_imageio/tool/download.py
```diff
@@ -0,0 +1,77 @@
...
+
+    cmd.add_argument(
+        "--max-workers",
+        type=options.Size(min_value=1, max_value=8),
+        default=client.MAX_WORKERS,
+        help=f"Maximum number of workers (default {client.MAX_WORKERS}).")
+
+    cmd.add_argument(
+        "--buffer-size",
+        type=options.Size(min_value=64 * KiB, max_value=16 * MiB),
```
rokkbert, line 42: It would be nice if the help text would include these limits, otherwise the user has to guess and retry until he's inside the legal interval.
Same for --max-workers, but the buffer-size is probably more variable.

ovirt_imageio/tool/options.py
```diff
@@ -0,0 +1,90 @@
...
+        cmd = self._commands.add_parser(name, help=help)
+        cmd.set_defaults(command=func)
+
+        cmd.add_argument(
+            "--engine-url",
+            help="ovirt-engine URL.")
+
+        cmd.add_argument(
+            "--username",
+            help="ovirt-engine username.")
```
rokkbert, line 37: The help shows this, --engine-url and --cafile) as optional. But are they really? The default username seems to be "None@N/A", which is unlikely to work, I assume. The default engine-url doesn't work either. Is there a default cafile?

bennyz: I suppose once --config/-c is implemented it will be really optional

rokkbert: Yes, that will be cool. I wonder if the parser is able to display that properly. Maybe like this:
`{--config CONFIG_FILE, --engine-url ENGINE_URL --username USERNAME --cafile CAFILE [--password-file PASSWORD_FILE]}`

ovirt_imageio/tool/download.py
nirs: This is unlikely to be a real issue since users are not expected to play with this, but it should be easy to expose the limits using the validator limits.

ovirt_imageio/tool/options.py
nirs: Yes once we have the config everything is optional, but we are missing validation step
after parsing arguments, since engine url/username/password are really required to connect to engine.

We don't have a way to show that an argument is required via command line or config file, and we don't want to write our own parser for this.

ovirt_imageio/client/_api.py
```diff
@@ -27,11 +27,14 @@
 from .. _internal.handlers import checksum as _checksum
 from .. _internal.nbd import UnixAddress
 
+BUFFER_SIZE = io.BUFFER_SIZE
+MAX_WORKERS = io.MAX_WORKERS
```
aesteve-rh, line 31: Why duplicate these constants?

nirs: This expose the constant in the API. This is exported to the package public names in __init__, and the code is using the _api module to access these values. This way to code
in the client does not have to import the io module.

nirs: Importing the constants does not make it clear that the names are part of the public module API.

for example:

    from .. _internal import io

Is "io" part of the _api public API?

Practically yes, you can use _api.io, but you should really not do this. This
makes your code fragile and it will break when this import will be replaced with:

    from .. _internal.io import suff

Defining the names in the module makes it clear that this is part of the module
API.

And doing here avoids the ugly internal import in __init__, and make exporting names
from the _api mdoule simpler and more consistent.

ovirt_imageio/tool/download.py
nirs: Done in current version.

ovirt_imageio/client/_api.py
aesteve-rh: Got it.

Maybe worth adding a comment (e.g., `# Exposed internal constants`) so that if we expose more internal constants in the public api they get all declared consecutively?

ovirt_imageio/client/_options.py
```diff
@@ -48,3 +50,45 @@ def parse(self):
...
+        return f"{n}{unit}"
+
+
+class Size:
+    """
+    Convert and validate size string.
+    """
+
+    def __init__(self, minimum=0, default=None, maximum=None):
+        self.minimum = None if minimum is None else SizeValue(minimum)
```
aesteve-rh, line 74: Is minimum `None` allowed? I ask because the value when omitted is `0`, and I guess a negative size does not make sense.

If you support None for consistency, maybe worth a test to ensure that it works and does not break in the future.

nirs: We can make minimum=None a hard error, since it does not make sense for a size. Sounds better than adding tests for non existing cases.

aesteve-rh: Sounds good, it also would save having to handle the case where `minimum is None` in the methods below :)

ovirt_imageio/client/_api.py
nirs: Added a comment.

ovirt_imageio/client/_options.py
nirs: mininimum=None raises now TypeError, unneeded check removed.

aesteve-rh (approved):
Looks great now!

ovirt_imageio/client/_ovirt.py
```diff
@@ -0,0 +1,378 @@
...
+    hosts = hosts_service.list(search=query, case_sensitive=True)
+    if len(hosts) == 0:
+        log.debug(
+            "Cannot use host with hardware id %s, host is not up, or does "
+            "not belong to data center %s",
+            vdsm_id, data_center.name)
+        return None
+
+    host = hosts[0]
+    log.debug("Using host id %s", host.id)
```
rokkbert, line 153: This isn't the hardware ID, where can I see that in the oVirt-UI?

ovirt_imageio/client/_ovirt.py
```diff
@@ -0,0 +1,378 @@
...
+            raise RuntimeError(f"Transfer {transfer.id} has failed")
+
+        if transfer.phase == types.ImageTransferPhase.PAUSED_SYSTEM:
+            transfer_service.cancel()
+            raise RuntimeError(f"Transfer {transfer.id} was paused by system")
+
+        if transfer.phase == types.ImageTransferPhase.TRANSFERRING:
+            break
+
+        if transfer.phase != types.ImageTransferPhase.INITIALIZING:
```
rokkbert, line 245: The PAUSED_* phases are not valid but are a reason to cancel?

ovirt_imageio/client/_ovirt.py
```diff
@@ -0,0 +1,378 @@
...
+def finalize_transfer(con, transfer, disk, timeout=300):
+    """
+    Finalize a transfer, making the transfer disk available.
+
+    If finalizing succeeds: the disk status will change to OK and
+    transfer's phase will change to FINISHED_SUCCESS.
+    On upload errors: the disk status will change to ILLEGAL, transfer's
+    phase will change to FINISHED_FAILURE and the disk will be removed.
+    In both cases the transfer entity:
+     a. prior to 4.4.7: removed shortly after the command finishes
```
rokkbert, line 289: Sounds better with "is" before the removed.

rokkbert (approved):
Nice, looks very clean and readable now!

ovirt_imageio/client/_ovirt.py
nirs: Yes this is the host id, displayed in engine UI in the host tab.
I think the hardware id is also displayed in the same tab.

ovirt_imageio/client/_ovirt.py
nirs: Yes, if the system paused the transfer something is wrong and we cannot continue.
The best way is to cancel the transfer and fail.

ovirt_imageio/client/_ovirt.py
nirs: Agree, will update in the next version.

nirs: Fixed in current version.

aesteve-rh (approved):

michalskrivanek (commented):
is it not packaged into rpm?

ovirt_imageio/client/_options.py
```diff
@@ -21,36 +23,75 @@ def __init__(self):
...
+
+    def _merge_config(self, args):
+        config = configparser.ConfigParser(interpolation=None)
+        config.read([self.config_file])
+
+        if not config.has_section(args.config):
+            self._parser.error(
+                f"No section: '{args.config}' in '{self.config_file}'")
+
+        for name in ("engine_url", "username", "password", "cafile"):
```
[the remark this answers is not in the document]
rokkbert: Ok, the "password/password-file"-thing seems to be a feature of the parser? Because it's the same for "engine/engine-url".

rokkbert (approved):

nirs:
> is it not packaged into rpm?

Not yet, it is not complete yet.

ovirt_imageio/client/_options.py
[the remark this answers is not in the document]
nirs: When you add --foo-bar the parser save the value in args.foo_bar. You can specify dest= if you want to make this explicit but this is well known.

password and --password-file are not the same - there are 3 ways to specify a password:
- --password-file: read password from file
- password in config file: read password from config file
- not set: read from stdin

ovirt_imageio/client/_options.py
```diff
@@ -21,36 +23,75 @@ def __init__(self):
...
+
         return args
 
+    @property
+    def config_file(self):
+        # https://specifications.freedesktop.org/basedir-spec
+        base_dir = os.environ.get("XDG_CONFIG_HOME")
+        if not base_dir:
+            base_dir = os.path.expanduser("~/.config")
+        return os.path.join(base_dir, "ovirt-img.conf")
```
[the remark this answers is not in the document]
nirs: ovirt.conf was used by the ovirt python sdk examples. This is the first time this code is packaged as a real tool so it makes sense to use the standard naming, which is the name of the tool. Users can copy or symlink ovirt.conf since both use the same format.

ovirt_imageio/client/_options.py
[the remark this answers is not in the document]
rokkbert: And there's this behaviour, which to me seems unspecified, where it is trying to open a password file that was given with --password:

`
[rob@host-vdsm ovirt-imageio]$ ./ovirt-img download-disk --password x --engine-url https://rob-ovirt-engine.local --username admin@internal --cafile /home/rob/cert.pem -f raw f931f3b6-669a-46d2-8309-f23d0575cdaf  diskimage 
  File "./ovirt-img", line 13, in <module>
    _tool.main()
  File "/home/rob/share/git/nir/ovirt-imageio/ovirt_imageio/client/_tool.py", line 26, in main
    args.command(args)
  File "/home/rob/share/git/nir/ovirt-imageio/ovirt_imageio/client/_download.py", line 62, in download_disk
    con = _ovirt.connect(args)
  File "/home/rob/share/git/nir/ovirt-imageio/ovirt_imageio/client/_ovirt.py", line 46, in connect
    with open(args.password_file) as f:
FileNotFoundError: [Errno 2] No such file or directory: 'x'
`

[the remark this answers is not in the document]
rokkbert: That behaviour might be "well known" to python-args-parser-programmers, but not necessarily to the users of this tool.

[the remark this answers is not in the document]
nirs: This is a feature of argparse (and geopt) - you can use partial names when setting arguments. All of these are the same:

    --password-file
    --password
    --pass
    --p

If we had a --password argument this would not work, but we don't.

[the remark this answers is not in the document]
rokkbert: Ok, to clarify: it seems to just automatically make short-options for the given long-options, as long as they are unique.

PR #104, opened 2022-07-24
Progress bar cleanups

nirs:
- Spelling
- Make private attributes private
- Remove wrong check
- Ensure closed progress is not updated
- Modernize ProgressBar._draw
- Make conditional expression more clear
- Use unit constants in ui tests

Rebased on top of #102.

### DIFF
--- a/ovirt_imageio/client/_ui.py
+++ b/ovirt_imageio/client/_ui.py
@@ -15,7 +15,7 @@ from .. _internal import util
 
 class ProgressBar:
 
-    def __init__(self, size=0, output=sys.stdout, step=None, width=79,
+    def __init__(self, size=None, output=sys.stdout, step=None, width=79,
                  now=time.monotonic):
         """
         Arguments:
@@ -29,16 +29,16 @@ class ProgressBar:
             now (callable): callable returning current time for testing.
         """
         self.size = size
-        self.output = output
+        self._output = output
         # TODO: use current terminal width instead.
-        self.width = width
-        self.now = now
-        self.lock = threading.Lock()
-        self.start = self.now()
+        self._width = width
+        self._now = now
+        self._lock = threading.Lock()
+        self._start = self._now()
         # Number of bytes transferred.
-        self.done = 0
+        self._done = 0
         # Value in percent. We start with -1 so the first update will show 0%.
-        self.value = -1
+        self._value = -1
 
         # The first update can take some time.
         self._draw()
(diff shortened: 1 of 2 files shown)


test/client_ui_test.py
```diff
@@ -35,59 +35,45 @@ def test_draw():
...
-    assert f.last == (
-        "[  50.00% ] 1.50 GiB, 1.20 seconds, 1.25 GiB/s".ljust(79) + "\r"
-    )
-
-    # More zeros, rates increases...
-    fake_time.now += 0.1
-    pb.update(1024**3)
+    # Write zeros (much faster)...
+    fake_time.now += 0.2
+    pb.update(2 * 1024**3)
```
aesteve-rh, line 60: This could use the recently added `units` constants.

ovirt_imageio/client/_ui.py
```diff
@@ -75,19 +75,11 @@ def close(self):
 
     def _draw(self, last=False):
         elapsed = self._now() - self._start
+        progress = f"{self._value:3d}%" if self.size else "----"
```
aesteve-rh, line 78: As size can now be `None`, maybe here we want to use `if self.size is not None` instead?

ovirt_imageio/client/_ui.py
```diff
@@ -15,87 +15,81 @@
...
-        self.start = self.now()
-        self.next = 0
-        self.done = 0
+        self._width = width
+        self._now = now
+        self._lock = threading.Lock()
+        self._start = self._now()
+        # Number of bytes transferred.
+        self._done = 0
+        # Value in percent. We start with -1 so the first update will show 0%.
```
rokkbert, line 40: But now we call _draw immediately, not update any more. So we will draw first with self._value -1, at least if size was specified. Is that what we want?

ovirt_imageio/client/_ui.py
```diff
@@ -15,87 +15,81 @@
...
-                self._draw(self.now(), last=True)
-
-    def _draw(self, now, last=False):
-        elapsed = now - self.start
-
-        if self.size:
-            progress = "%6.2f%%" % (self.done / self.size * 100)
-        else:
-            progress = "-------"
+        with self._lock:
```
rokkbert, line 71: Cool that it can be written like this :)

ovirt_imageio/client/_ui.py
nirs: But caller can use size=0, and this will try to divide by zero.

ovirt_imageio/client/_ui.py
nirs: This does not happen in current code since we never set size when creating the progress bar, but it is a bug. We also miss a test for setting size upfront.

aesteve-rh (approved):

ovirt_imageio/client/_ui.py
nirs: Fixed in master now.

aesteve-rh (approved):

Issue #72, opened 2022-05-24
ovirt-img command line tool


ahadas:
Uploading an image is complex process, requiring:
- detecting the image type (qcow2, raw, iso)
- creating a disk using the SDK, with the right format, sparse, initial_size
  and provional_size, which depend on the storage domain type
- checking if the current host can be used for the transfer
- starting a transfer on the current host if possible
- if the transfer be started on the current host, the user have to choose the
  transfer_url or the proxy_url.
- uploading the file using imageio REST API.
- cancel transfer on failures

We have example code and documentation for some of the steps, but building
a correct upload application is too complicated. Even virt-v2v did not get it 
right yet, after few month of development, with help form the oVirt team.

Uploading disks with several snapshots is even more complex.

It does not make sense that all users will have to write this code and make the
same errors.

The upload and download examples in the SDK are usually wrong, outdated, and do
not use the latest features offered by imageio. They do not have any automated
tests so they are likely to break after every update.

The upload and download examples also contain huge amount of duplicate code, and
require too much time to maintain.

Finally the ansible modules are using the outdated example code and need extra
maintenance and testing.

All these issues can be solved by providing a command line tool that does
everything for you.

Something like:

    ovirt-img upload --storage my-storage my-image.qcow2 https://example.com/

The command will:
- read engine username and password from standard locations in a secure way
- inspect the image properties and create disk property, matching the storage
  domain capabilities to the image type
- inspect the current host and start the upload on it if posible
- if the transfer cannot be started on current host, select the best transfer
  destination
- upload the data efficiently, supporting sparse images
- handle failures properly, avoiding leaving paused transfers
- support resuming if the upload was aborted
- show proper progress during upload

The data path part is already implemented in https://gerrit.ovirt.org/c/94216/

Original bug: https://bugzilla.redhat.com/1626262

nirs:
First version completed by #116. We will open new issues for additional commands.

PR #105, opened 2022-07-25
ovirt-img: Add logging options

nirs:
More general options parsing allowing mixing of command line
arguments and config options. Add log file options using the
new infrastructure.

This is part of #72

### DIFF
--- a/ovirt_imageio/client/_options.py
+++ b/ovirt_imageio/client/_options.py
@@ -20,6 +20,29 @@ import os
 import sys
 
 
+class Choices:
+
+    def __init__(self, name, values):
+        self.name = name
+        self.values = values
+
+    def __call__(self, s):
+        if s not in self.values:
+            raise ValueError(
+                f"Invalid '{self.name}' value: '{s}', choices: {self}")
+        return s
+
+    def __str__(self):
+        s = ", ".join(self.values)
+        return f"{{{s}}}"
+
+    def __repr__(self):
+        return repr(self.name)
+
+
+log_level = Choices("log_level", ("debug", "info", "warning", "error"))
+
+
 class Option(
         namedtuple("Option", "name,args,config,required,type,default,help")):
 
@@ -88,6 +111,21 @@ class Parser:
             config=True,
             help=("Path to ovirt-engine CA certificate. If not set, read "
                   "from the specified config section"),
+        ),
+        Option(
+            name="log_file",
+            args=["--log-file"],
+            config=True,
+            default="/dev/stderr",
+            help="Log file name (default: /dev/stderr).",
+        ),
+        Option(
+            name="log_level",
+            args=["--log-level"],
+            config=True,
+            type=log_level,
+            default="warning",
+            help=(f"Log level (choices: {log_level}, default: warning)."),
         ),
     ]
 
--- a/ovirt_imageio/client/_options.py
+++ b/ovirt_imageio/client/_options.py
@@ -10,6 +10,8 @@
 Tool options.
 """
 
+from collections import namedtuple
+
 from .. _internal.units import KiB, MiB, GiB, TiB
 
 import argparse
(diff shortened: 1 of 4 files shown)


ovirt_imageio/client/_options.py
```diff
@@ -64,6 +70,18 @@ class Parser:
...
+            args=["--log-file"],
+            name="log_file",
+            help="If set, log to file.",
+        ),
+        Option(
+            args=["--log-level"],
+            name="log_level",
+            type=log_level,
+            default="info",
+            help="Logging level (default: info).",
```
aesteve-rh, line 83: As an alternative, https://docs.python.org/3/library/argparse.html#argparse.ArgumentDefaultsHelpFormatter allows to have the default value added automatically to the help messages.

nirs: It does not work for us, since we must define all options *without* a default. If we use a default, after parsing arguements, all arguments have a default, so when we come to merge the config values we cannot detect an argument which was not set.

aesteve-rh (approved):
LGTM :)

nirs:
@rokkbert do you have time to review it now?

ovirt_imageio/client/_options.py
```diff
@@ -10,62 +10,124 @@
...
 from .. _internal.units import KiB, MiB, GiB, TiB
 
 import argparse
 import configparser
 import os
+import sys
+
+
+class Option(
+        namedtuple("Option", "name,args,config,required,type,default,help")):
```
rokkbert, line 24: I think it would be good to have a comment here, to describe what the fields mean, esp. "args" and "config".

ovirt_imageio/client/_options.py
```diff
@@ -10,62 +10,124 @@
...
+        ),
+        Option(
+            name="password",
+            config=True,
+        ),
+        Option(
+            name="password_file",
+            args=["--password-file"],
+            help=("Read ovirt-engine password from file. If not set, read "
+                  "password from the specified config section."),
```
rokkbert, line 72: "... or prompt user."?

ovirt_imageio/client/_options.py
```diff
@@ -10,62 +10,124 @@
...
+            config=True,
+            help="If set, log to file.",
+        ),
+        Option(
+            name="log_level",
+            args=["--log-level"],
+            config=True,
+            type=log_level,
+            default="info",
+            help="Logging level (default: info).",
```
rokkbert, line 93: Should the levels be listed in the help? Some systems have more/different levels.

ovirt_imageio/client/_options.py
```diff
@@ -78,19 +140,54 @@ def config_file(self):
...
         config = configparser.ConfigParser(interpolation=None)
         config.read([self.config_file])
 
         if not config.has_section(args.config):
             self._parser.error(
                 f"No section: '{args.config}' in '{self.config_file}'")
 
-        for name in ("engine_url", "username", "password", "cafile"):
-            if getattr(args, name) is not None:
+        for option in self._OPTIONS:
```
rokkbert, line 155: That's a nice and clear way, to fill in the config-file-arguments after the CLI-arguments!

ovirt_imageio/client/_options.py
```diff
@@ -78,19 +140,54 @@ def config_file(self):
...
+                setattr(args, option.name, option.default)
+
+    def _check_required_options(self, args):
+        """
+        Check that all required options have some value. This must be done
+        after setting the defaults.
+        """
+        for option in self._OPTIONS:
+            if option.required and getattr(args, option.name) is None:
+                self._parser.error(f"Option '{option.name}' is required")
```
rokkbert, line 190: It's suboptimal that we have 2 levels of required-arguments-checking, instead of telling the user the whole list:

```
$ ./ovirt-img  download-disk 931f3b6-669a-46d2-8309-f23d0575cdaf 
usage: ovirt-img download-disk [-h] [-c CONFIG] [--engine-url ENGINE_URL]
                               [--username USERNAME]
                               [--password-file PASSWORD_FILE]
                               [--cafile CAFILE] [--log-file LOG_FILE]
                               [--log-level LOG_LEVEL] [-f {raw,qcow2}]
                               [--max-workers MAX_WORKERS]
                               [--buffer-size BUFFER_SIZE]
                               disk_id filename
ovirt-img download-disk: error: the following arguments are required: filename
$ ./ovirt-img  download-disk 931f3b6-669a-46d2-8309-f23d0575cdaf diskname
usage: ovirt-img [-h] {download-disk} ...
ovirt-img: error: Option 'engine_url' is required
```

But I think it's ok, he can do it step by step too.

ovirt_imageio/client/_options.py
rokkbert, line 93: Since log_level without log_file (which has no default) is useless, should that combination be an error? Or should we log to stdout?

rokkbert (dismissed):
I would expand the help and comments a bit, but in general it's a good improvement.

ovirt_imageio/client/_options.py
nirs: Yes, we need to list the levels here.

ovirt_imageio/client/_options.py
nirs: Right, it should be easy to fail with the all the missing arguments instead.

ovirt_imageio/client/_options.py
nirs: We set the log level only when the log file is set, but actually the logger is using filename=stderr and level=warning by default. I think we can make this more general by setting the log level in all case. If you use --log-level=debug your will get very verbose output in the shell.

ovirt_imageio/client/_options.py
nirs: The arguments are documented now.

ovirt_imageio/client/_options.py
nirs: Added.

ovirt_imageio/client/_options.py
nirs: The choices are listed now.

ovirt_imageio/client/_options.py
nirs: Fixing the 2 levels of required arguments checking may be possible if we teach argparse to make disk_id and filename optional, and implement the check in our parser. But I don't have time for this now, we can improve this later based on feedback from users.

We show now all the missing required arguments instead of the first one.

ovirt_imageio/client/_options.py
nirs: Log level is always used now, and the default value is warning.

Example usage with default log level:
```
$ ./ovirt-img download-disk -c engine adb9b2c7-32a9-4e87-894c-710de7165086 \
    /data/fc/tmp/dl.qcow2
[ 100% ] 6.00 GiB, 1.85 seconds, 3.24 GiB/s                                    
```

With info log level:
```
$ ./ovirt-img download-disk -c engine adb9b2c7-32a9-4e87-894c-710de7165086 \
    /data/fc/tmp/dl.qcow2 --log-level info
2022-07-27 01:19:00,643 INFO    (MainThread) [ovirt] Creating transfer image=Disk(id='adb9b2c7-32a9-4e87-894c-710de7165086') direction=download host=None backup=None shallow=None timeout_policy=cancel
2022-07-27 01:19:01,205 INFO    (MainThread) [ovirt] Transfer 'ed600506-61ed-4bfe-afb9-76662e7cd551' ready on host 'host4' in 0.6 seconds
[ 100% ] 6.00 GiB, 1.86 seconds, 3.22 GiB/s                                    
2022-07-27 01:19:03,067 INFO    (MainThread) [ovirt] Finalizing transfer 'ed600506-61ed-4bfe-afb9-76662e7cd551'
2022-07-27 01:19:05,103 INFO    (MainThread) [ovirt] Transfer 'ed600506-61ed-4bfe-afb9-76662e7cd551' finalized in 2.0 seconds
```

aesteve-rh (approved):

ovirt_imageio/client/_options.py
```diff
@@ -10,62 +10,159 @@
...
+            args (List[str]): If set, add command line arguments for this
+                option.
+            config (bool): If True this option can be set in the config file.
+            required (bool): If True this option is required and parsing
+                arguments will fail if it is not specified in the command line
+                or config file.
+            type (callable): Callable converting the command line argument or
+                config value to the wanted type. Must raise ValueError if the
+                value is invalid and cannot be converted.
+            default (Any): The default value if the option was not set int the
```
rokkbert, line 65: typo "int"

rokkbert (approved):
Good documentation update!

ovirt_imageio/client/_options.py
nirs: Fixed in current version.

aesteve-rh (approved):

PR #107, opened 2022-07-27
ovirt-img: Terminate cleanly on signals

nirs:
Handle SIGINT and SIGTERM by terminating qemu-nbd and triggering normal cleanup flow, removing the temporary directory and finalizing the image transfer.

When qemu-nbd is terminated, the worker threads may log noisy and unhelpful tracebacks. Since we already re-raise the first worker error, we can suppress these tracebacks.

When the tool is terminated by signal, we suppress errors from the tool (e.g. BrokenPipe). Otherwise we log one traceback on errors.

Here are some examples:

Terminated by SIGINT:
```
$ ./ovirt-img download-disk -c engine 07b53997-7b4c-4861-bf38-f6fbb29921d9 dl.qcow2
^C[   5% ] 2.50 GiB, 3.66 seconds, 698.52 MiB/s                                  
2022-07-27 14:29:58,892 ERROR   (MainThread) [tool] Terminated by signal 2
```

qemu-nbd killed during transfer:
```
$ ./ovirt-img download-disk -c engine 07b53997-7b4c-4861-bf38-f6fbb29921d9 dl.qcow2
[  13% ] 6.75 GiB, 8.88 seconds, 778.06 MiB/s                                  
2022-07-27 14:30:53,638 ERROR   (MainThread) [tool] Command failed
  File "/home/nsoffer/src/ovirt-imageio/ovirt_imageio/client/_tool.py", line 41, in main
    args.command(args)
  File "/home/nsoffer/src/ovirt-imageio/ovirt_imageio/client/_download.py", line 72, in download_disk
    _api.download(
  File "/home/nsoffer/src/ovirt-imageio/ovirt_imageio/client/_api.py", line 176, in download
    io.copy(
  File "/home/nsoffer/src/ovirt-imageio/ovirt_imageio/_internal/io.py", line 44, in copy
    with Executor(name=name) as executor:
  File "/home/nsoffer/src/ovirt-imageio/ovirt_imageio/_internal/io.py", line 196, in __exit__
    self.stop()
  File "/home/nsoffer/src/ovirt-imageio/ovirt_imageio/_internal/io.py", line 173, in stop
    raise self._errors[0]
  File "/home/nsoffer/src/ovirt-imageio/ovirt_imageio/_internal/io.py", line 245, in _run
    handler.copy(req)
  File "/home/nsoffer/src/ovirt-imageio/ovirt_imageio/_internal/io.py", line 289, in copy
    self._src.write_to(self._dst, req.length, self._buf)
  File "/home/nsoffer/src/ovirt-imageio/ovirt_imageio/_internal/backends/http.py", line 220, in write_to
    writer.write(view[:n])
  File "/home/nsoffer/src/ovirt-imageio/ovirt_imageio/_internal/backends/nbd.py", line 119, in write
    self._client.write(self._position, buf)
  File "/home/nsoffer/src/ovirt-imageio/ovirt_imageio/_internal/nbd.py", line 444, in write
    self._send(data)
  File "/home/nsoffer/src/ovirt-imageio/ovirt_imageio/_internal/nbd.py", line 1227, in _send
    self._sock.sendall(data)
BrokenPipeError: [Errno 32] Broken pipe
```

### DIFF
--- a/ovirt_imageio/client/_api.py
+++ b/ovirt_imageio/client/_api.py
@@ -20,16 +20,17 @@ from contextlib import contextmanager
 from urllib.parse import urlparse
 
 from .. _internal import blkhash
-from .. _internal import io
 from .. _internal import qemu_img
 from .. _internal import qemu_nbd
 from .. _internal.backends import http, nbd
 from .. _internal.handlers import checksum as _checksum
 from .. _internal.nbd import UnixAddress
 
+from . import _io
+
 # API constants.
-BUFFER_SIZE = io.BUFFER_SIZE
-MAX_WORKERS = io.MAX_WORKERS
+BUFFER_SIZE = _io.BUFFER_SIZE
+MAX_WORKERS = _io.MAX_WORKERS
 
 log = logging.getLogger("client")
 
@@ -95,7 +96,7 @@ def upload(filename, url, cafile, buffer_size=BUFFER_SIZE, secure=True,
                 backing_chain=backing_chain) as src:
 
             # Upload the image to the server.
-            io.copy(
+            _io.copy(
                 src,
                 dst,
                 max_workers=max_workers,
@@ -173,7 +174,7 @@ def download(url, filename, cafile, fmt="qcow2", incremental=False,
         with _open_nbd(filename, fmt, shared=max_workers) as dst:
 
             # Download the image from the server to the local image.
-            io.copy(
+            _io.copy(
                 src,
                 dst,
                 dirty=incremental,
--- a/ovirt_imageio/client/_tool.py
+++ b/ovirt_imageio/client/_tool.py
@@ -11,9 +11,16 @@ Tool for transferring disk images.
 """
 
 import logging
+import os
+import signal
+import sys
 
 from . import _options
 from . import _download
+
+log = logging.getLogger("tool")
+
+terminated = None
 
 
 def main():
@@ -27,7 +34,23 @@ def main():
         format="%(asctime)s %(levelname)-7s (%(threadName)s) [%(name)s] "
                "%(message)s")
 
-    # XXX Setup signal handlers.
-    # XXX Handle commands errors.
+    signal.signal(signal.SIGINT, _handle_signal)
+    signal.signal(signal.SIGTERM, _handle_signal)
 
-    args.command(args)
+    try:
+        args.command(args)
+    except Exception:
+        if terminated:
+            log.error("Terminated by signal %d", terminated)
+            sys.exit(128 + terminated)
+        else:
+            log.exception("Command failed")
+            sys.exit(1)
+
+
+def _handle_signal(signo, frame):
+    global terminated
+    if not terminated:
+        terminated = signo
+        # Terminate qemu-nbd triggering normal cleanup flow.
+        os.killpg(0, signo)
--- a/ovirt_imageio/_internal/io.py
+++ b/ovirt_imageio/_internal/io.py
@@ -251,7 +251,7 @@ class Worker:
         except Exception as e:
             self._errors.append(e)
             self._queue.close()
-            log.exception("Worker %s failed", self._name)
+            log.debug("Worker %s failed: %s", self._name, e)
         else:
             log.debug("Worker %s finished", self._name)
 
--- a/ovirt_imageio/client/_io.py
+++ b/ovirt_imageio/client/_io.py
@@ -16,9 +16,9 @@ from collections import deque, namedtuple
 from contextlib import closing
 from functools import partial
 
-from . import util
-from . backends import Wrapper
-from .units import MiB
+from .. _internal import util
+from .. _internal.backends import Wrapper
+from .. _internal.units import MiB
 
 # Limit maximum zero and copy size to spread the workload better to multiple
 # workers and ensure frequent progress updates when handling large extents.
--- a/test/client_io_test.py
+++ b/test/client_io_test.py
@@ -12,11 +12,12 @@ import pytest
 from urllib.parse import urlparse
 
 from ovirt_imageio._internal import extent
-from ovirt_imageio._internal import io
 from ovirt_imageio._internal import qemu_img
 from ovirt_imageio._internal import qemu_nbd
 from ovirt_imageio._internal.backends import nbd, memory
 from ovirt_imageio._internal.nbd import UnixAddress
+
+from ovirt_imageio.client import _io
 
 ZERO_PARAMS = [
     # Copying to image with unknown content.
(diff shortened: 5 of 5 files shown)


ovirt_imageio/client/_tool.py
```diff
@@ -27,7 +34,23 @@ def main():
...
-    # XXX Setup signal handlers.
-    # XXX Handle commands errors.
-
-    args.command(args)
+    signal.signal(signal.SIGINT, terminate)
+    signal.signal(signal.SIGTERM, terminate)
+
+    try:
+        args.command(args)
+    except Exception:
```
aesteve-rh, line 42: Catching `Exception` is necessary, but there are cases where it may be nice to catch specific exceptions.

`ovirtsdk4.Error` is a good candidate to be catch separately. We could do `log.error` with the exception string instead of `log.exception`, so we skip the Traceback. Traceback is useful for developers, but probably just noisy for users.

nirs: I don't want to depend on sdk types here. If we think that dome errors need special handling, we need to wrap the sdk error with tool specific errors.

We can add some standard errors like client._errors.NoSuchDisk, client._errors.NoSuchStorageDomain, client._errors.NoSuchImage and show a traceback only for unexpected errors.

Lets open an issue for this for now. First version always logging tracebacks is OK.

nirs: Opened #112 to track this. This may not be easy to fix in all cases.

aesteve-rh (changes requested):
LGTM, only a couple minor comments.

ovirt_imageio/client/_tool.py
```diff
@@ -27,7 +34,23 @@ def main():
...
+            log.exception("Command failed")
+            sys.exit(1)
+
+
+def terminate(signo, frame):
+    global terminated
+    if not terminated:
+        terminated = signo
+        # Terminate qemu-nbd triggering normal cleanup flow.
+        os.killpg(0, signo)
```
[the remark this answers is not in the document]
nirs: You cannot do this from signal handler, and tools should not do this.

ovirt_imageio/client/_tool.py
```diff
@@ -27,7 +34,23 @@ def main():
...
+    except Exception:
+        if terminated:
+            log.error("Terminated by signal %d", terminated)
+            sys.exit(128 + terminated)
+        else:
+            log.exception("Command failed")
+            sys.exit(1)
+
+
+def terminate(signo, frame):
```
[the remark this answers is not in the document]
nirs: 👍

ovirt_imageio/client/_tool.py
rokkbert, line 56: Is the 0 here a magic shorthand for "my own process group id"? I saw a lot of examples where people explicitly find the id first.

nirs: 0 is not really magic, it means no signals was delivered (since signal number starts at 1) and it is also palsy value.

nirs: Changing the initial value to None to avoid this confusion.

rokkbert:
The useless stacktraces are gone, but I'm not sure which temporary files need removal. Without this patch, if I interrupt with ^C, the directory /tmp/imageio-* already gets removed. Are there more? Or am I lucky, and there are situations when it doesn't?

ovirt_imageio/client/_tool.py
rokkbert: But the first argument to killpg is the process-group-ID.
If I try it in a test program it does kill itself, so I assume it must mean "my group". It just doesn't seem to be very common to use the 0.

nirs:
> The useless stacktraces are gone, but I'm not sure which temporary files need removal. Without this patch, if I interrupt with ^C, the directory /tmp/imageio-* already gets removed. Are there more? Or am I lucky, and there are situations when it doesn't?

When interrupting the tool (Control+C) with current code there is no
issue since python convert SIGINT to KeyboardInterrupt exception, and
the exception is handled by context managers and try-finally block in
_download.download_disk.

But when terminating the tool with SIGTERM (kill {pid}), the tool exits
without any cleanup:

- qemu-nbd continue to run forever
- /tmp/imageio-xxx not removed
- the transfer is not finalized (engine will cancel it in 60 seconds)
- the connection to engine is not closed (bad for engine auth code)

With this change the flow is:

1. Receiving termination signal (SIGINT, SIGTERM)
2. Kill qemu-nbd
3. _io.copy() detects a failed worker, stops all workers and raises an error from the first failing worker
4. _api.download() terminates qemu-nbd, removes the temporary directory and raises the error
5. _download.download_disk() finalizes the transfer, closes the connection and raises error.
6.  _tool.main() logs an error

This is the same flow handling qemu-nbd errors or imageio server errors,
but _tool.main() knows that the tool was not interrupted so it will log
the error raised from _download.download_disk().

ovirt_imageio/client/_tool.py
```diff
@@ -27,7 +34,23 @@ def main():
...
-    # XXX Handle commands errors.
-
-    args.command(args)
+    signal.signal(signal.SIGINT, _handle_signal)
+    signal.signal(signal.SIGTERM, _handle_signal)
+
+    try:
+        args.command(args)
+    except Exception:
+        if terminated:
```
aesteve-rh, line 43: As it is right now, terminated will never be `0`, so this is fine, but technically it should be `if terminated is not None:`. Similar but opposite in the handler.

aesteve-rh (dismissed):
Looks good. Small comment below, but feel free to ignore it since it is harmless.

rokkbert (dismissed):

rokkbert (approved):
Works.

ovirt_imageio/client/_tool.py
nirs: Thinking again, since signal cannot be 0, using terminated as boolean is much more clear and better then using "is not None".

ovirt_imageio/client/_api.py
```diff
@@ -20,16 +20,17 @@
...
 from .. _internal.handlers import checksum as _checksum
 from .. _internal.nbd import UnixAddress
 
+from . import _io
+
 # API constants.
-BUFFER_SIZE = io.BUFFER_SIZE
-MAX_WORKERS = io.MAX_WORKERS
+BUFFER_SIZE = _io.BUFFER_SIZE
+MAX_WORKERS = _io.MAX_WORKERS
```
aesteve-rh, line 33: With the change these duplicates can be omitted.

nirs: Why, we still export these names

aesteve-rh: Uhm, because they can now be exported directly from `_io` in the `__init__` module. But I guess the intend is to have the `_api` to expose the names.

It's fine.

aesteve-rh (approved):

PR #110, opened 2022-07-27
client: Show progress phase

nirs:
Transferring an image include several phases that may take several
seconds, and we want to show what the tool is doing. When we finish, we
want to show that the operation was completed. If we failed, we want to
show a short failure message even if we use log errors to file.

Add an optional phase argument to ProgressBar. If the phase is set, it
is displayed at the end:

    [ ---- ] 0 bytes, 0.00 s, 0 bytes/s | setting up

The caller can change the phase by setting a new value:

    pb.phase = "downloading image"

This will redraw the progress bar:

    [   0% ] 0 bytes, 0.00 s, 0 bytes/s | downloading image

This change only adds the infrastructure, no change yet in the actual
commands.

### DIFF
--- a/test/client_ui_test.py
+++ b/test/client_ui_test.py
@@ -37,45 +37,39 @@ def test_draw():
 
     # Size is unknown at this point.
     pb = client.ProgressBar(output=f, now=fake_time)
-    assert f.last == (
-        "[ ---- ] 0 bytes, 0.00 seconds, 0 bytes/s".ljust(79) + "\r"
-    )
+    line = "[ ---- ] 0 bytes, 0.00 seconds, 0 bytes/s"
+    assert f.last == line.ljust(79) + "\r"
 
     # Size was updated, but no bytes were transfered yet.
     fake_time.now += 0.1
     pb.size = 3 * GiB
     pb.update(0)
-    assert f.last == (
-        "[   0% ] 0 bytes, 0.10 seconds, 0 bytes/s".ljust(79) + "\r"
-    )
+    line = "[   0% ] 0 bytes, 0.10 seconds, 0 bytes/s"
+    assert f.last == line.ljust(79) + "\r"
 
     # Write some data...
     fake_time.now += 1.0
     pb.update(512 * MiB)
-    assert f.last == (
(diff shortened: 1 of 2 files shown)


test/client_ui_test.py
```diff
@@ -94,21 +113,57 @@ def test_close():
...
+        phase="starting operation",
+        error_phase="operation failed",
+        size=GiB,
+        output=f)
+    with pytest.raises(RuntimeError):
+        with pb:
+            raise RuntimeError
+
+    assert f.last.endswith("\n")
+    assert pb.phase == "operation failed"
```
rokkbert, line 169: When reading the code I was wondering, if/how this error-phase actually gets drawn. I saw it then, but maybe the test should verify that too, so it's not just some internal state.

rokkbert (approved):
LGTM, but I would expand the tests to check that the error_phase actually does get displayed.

test/client_ui_test.py
nirs: Fixed in current version, we test also the output.

test/client_ui_test.py
```diff
@@ -37,55 +37,48 @@ def test_draw():
...
 def test_with_size():
     fake_time = FakeTime()
     f = FakeFile()
 
     client.ProgressBar(size=3 * GiB, output=f, now=fake_time)
-    assert f.last == (
-        "[   0% ] 0 bytes, 0.00 seconds, 0 bytes/s".ljust(79) + "\r"
-    )
+    line = "[   0% ] 0 bytes, 0.00 seconds, 0 bytes/s"
+    assert f.last == line.ljust(79) + "\r"
```
aesteve-rh, line 107: Since this assertions all look the same, maybe worth extracting them to a function?
```python
def assert_last_progress_match(file, expected_line):
    assert file.last == expected_line.ljust(79) + "\r"
```

aesteve-rh (approved):
Other than the comment, all good. Nice enhancement!

test/client_ui_test.py
nirs: These are not new, let’s improve this later.

Issue #121, opened 2022-07-29
Add option for custom disk alias for uploaded disk


aesteve-rh:
ovirt-img upload-disk command assigns a fixed disk alias based on the source image name.

Add a command line option to allow adding custom disk aliases.

nirs:
This is very useful - changing iso disk name is not supported yet, so the only way to set the name is before the upload. It should be fixed in engine but it may not be easy. Adding an option is trivial.

Adding to 4.5.2 since this should be very easy.

PR #122, opened 2022-08-01
ovirt-img: add name option for disk

aesteve-rh:
Add '--name' option to the 'upload-disk' command
to allow users to define a custom alias name
for the new uploaded disk. If ommited, default
behaviour remains, taking the source image filename
as an alias.
```
usage: ovirt-img upload-disk [-h] [-c CONFIG] [--engine-url ENGINE_URL]
                             [--username USERNAME]
                             [--password-file PASSWORD_FILE] [--cafile CAFILE]
                             [--log-file LOG_FILE] [--log-level LOG_LEVEL]
                             [--max-workers MAX_WORKERS]
                             [--buffer-size BUFFER_SIZE] -s STORAGE_DOMAIN
                             [-f {raw,qcow2}] [--preallocated]
                             [--disk-id DISK_ID] [--name NAME]
                             filename
....
  --name NAME           Alias name for the new disk. If not specified, name
                        will correspond with the image filename.
```
Closes: #121

### DIFF
--- a/ovirt_imageio/client/_upload.py
+++ b/ovirt_imageio/client/_upload.py
@@ -58,8 +58,13 @@ def register(parser):
     cmd.add_argument(
         "--disk-id",
         type=_options.UUID,
-        help="A UUID for the new disk. If not specified oVirt will "
+        help="A UUID for the new disk. If not specified, oVirt will "
              "create a new UUID.")
+
+    cmd.add_argument(
+        "--name",
+        help="Alias name for the new disk. If not specified, name will "
+             "correspond with the image filename.")
 
     cmd.add_argument(
         "filename",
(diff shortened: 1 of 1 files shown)


ovirt_imageio/client/_upload.py
```diff
@@ -129,7 +134,7 @@ def _prepare(args):
     name = os.path.splitext(os.path.basename(img_info["filename"]))[0]
```
nirs, line 134: If user specified a namme, we don't need to extract the name from the file name, so better do:

    if args.name is None:
        name = os.path.splitext(os.path.basename(img_info["filename"]))[0]
    else:
        name = args.name

nirs (changes requested):

nirs (approved):

Issue #76, opened 2022-05-25
Uploading huge sparse images is too slow


nirs:
Uploading 8T fedora image containing 1.5G of data takes 5.5 minutes.
It should be much faster.

## Client log

```
$ python3 /usr/share/doc/python3-ovirt-engine-sdk4/examples/upload_disk.py -c engine --sd-name fc-01 --disk-format qcow2 --disk-sparse fedora-35-8t.raw 
[   0.0 ] Checking image...
[   0.0 ] Image format: raw
[   0.0 ] Disk format: cow
[   0.0 ] Disk content type: data
[   0.0 ] Disk provisioned size: 8796093022208
[   0.0 ] Disk initial size: 3228303360
[   0.0 ] Disk name: fedora-35-8t.qcow2
[   0.0 ] Disk backup: False
[   0.0 ] Connecting...
[   0.0 ] Creating disk...
[   5.7 ] Disk ID: bdce52a0-d2ac-4391-81ba-88566f323a0d
[   5.7 ] Creating image transfer...
[   7.4 ] Transfer ID: b79eaba6-b633-4f54-bf27-1d206046ac96
[   7.4 ] Transfer host name: host4
[   7.4 ] Uploading image...
[ 100.00% ] 8.00 TiB, 323.11 seconds, 25.35 GiB/s                              
[ 330.5 ] Finalizing image transfer...
[ 333.6 ] Upload completed successfully
```

## Server log

```
2022-05-25 17:11:17,585 INFO    (Thread-1157) [tickets] [local] ADD transfer={'dirty': False, 'ops': ['write'], 'size': 8796093022208, 'sparse': False, 'inactivity_timeout': 60, 'transfer_id': 'b79eaba6-b633-4f54-bf27-1d206046ac96', 'uuid': '3f677555-3f2
9-47fc-beea-5e6339370a9c', 'timeout': 300, 'url': 'nbd:unix:/run/vdsm/nbd/b79eaba6-b633-4f54-bf27-1d206046ac96.sock'}
2022-05-25 17:11:17,588 INFO    (Thread-1157) [http] CLOSE connection=1157 client=local [connection 1 ops, 0.004801 s] [dispatch 1 ops, 0.003341 s]
2022-05-25 17:11:18,671 INFO    (Thread-1158) [http] OPEN connection=1158 client=::ffff:192.168.122.23
[...]

PR #123, opened 2022-08-01
Disk is zero

nirs:
When we upload to a new disk using qcow2 format or new file we know that
the disk is empty and the entire disk contents is read as zeros by the
guest. In this case we can skip zero extents on the source image instead
of zeroing them on the destination image.

Zeroing on the destination image is usually fast, but when uploading big
images it can slow down the upload a lot. When uploading to qcow2 image,
zeroing creates suboptimal image that will be slower to read and copy
later.

Example upload of empty 8 TiB image:

Before:

    ./ovirt-img upload-disk -c engine --storage-domain fc-01 empty-8t.qcow2
    [ 100% ] 8.00 TiB, 222.82 s, 36.77 GiB/s | upload completed

After:

    ./ovirt-img upload-disk -c engine --storage-domain fc-01 empty-8t.qcow2
    [ 100% ] 8.00 TiB, 11.51 s, 711.88 GiB/s | upload completed

Example upload of 8 TiB Fedora 35 image:

Before:

    $ ./ovirt-img upload-disk -c engine --storage-domain fc-01 fedora-35-8t.qcow2
    [ 100% ] 8.00 TiB, 317.88 s, 25.77 GiB/s | upload completed

After:

    $ ./ovirt-img upload-disk -c engine --storage-domain fc-01 fedora-35-8t.qcow2
    [ 100% ] 8.00 TiB, 109.13 s, 75.07 GiB/s | upload completed

Fixes: https://github.com/oVirt/ovirt-imageio/issues/76


ovirt_imageio/client/_upload.py
```diff
@@ -22,7 +22,7 @@
 
 DiskInfo = namedtuple(
     "DiskInfo",
-    "name,initial_size,provisioned_size,content_type,format,sparse")
+    "name,initial_size,provisioned_size,content_type,format,sparse,is_zero")
```
aesteve-rh, line 25: An idea not for this PR: Could this `namedtuple` maybe benefit of being wrapped into a class?

For example, `is_zero` does not need to be set, can be a class property:
```python
    @property
    def is_zero(self):
        return (self.format == FORMAT_QCOW2) or self.sparse
```

All attributes could be stored as strings, and transformed into `sdk` types in `@property` getters, so we have a central point where this transformation is done.

ovirt_imageio/client/_api.py
```diff
@@ -69,6 +69,10 @@ def upload(filename, url, cafile, buffer_size=BUFFER_SIZE, secure=True,
             image data, leaving unallocated areas as holes, exposing data from
             the target disk backing chain. Valid only when uploding to an empty
             snapshot.
+        disk_is_zero (bool): If set, skip zero extents in the source image
+            instead of zeroing the extent in the destination disk. Should be
+            used only when uploading to new disk on file based storage or new
+            sparse disk on block based storage.
```
aesteve-rh, line 75: **At first glance** is a bit weird to read the parameter name, and then do `zero=not (disk_is_zero [...]` below. I do understand the logic: if the full disk is zero, then we can skip zero extents, so we do not need to zero the area on destination. But maybe another name would avoid confusion? What about `disk_is_empty`?

Also, I am curious about what would happen if we actually skip zero extents when uploading to non-empty disks? Will the disk be corrupted? Will it work but the disks would not be the same when compared?

test/client_test.py
```diff
@@ -299,6 +349,7 @@ def test_upload_shallow(srv, nbd_server, tmpdir, base_fmt):
         src_base,
         url,
         srv.config.tls.ca_file,
+        disk_is_zero=disk_is_zero,
```
aesteve-rh, line 352: But this is uploading a disk that is not zero, and this parameter can be True.
Wouldn't that be wrong usage according to the docstring?
```
Should used only when uploading to new disk on file based storage or new
sparse disk on block based storage.
```

aesteve-rh (commented):
The change looks good, and the speedup is awesome!

A few comments below, that are mostly clarifications.

ovirt_imageio/client/_upload.py
nirs: We can but we need to document why. We can move the comment to the property, but it means that the comment related to prepare are spread in different locations in the file. It better that all the logic is in one place.

ovirt_imageio/client/_api.py
nirs: disk_is_empty is not correct for raw preallocated disk on block storage - the disk is empty, but when we read from it we may get data from previous use of the device. It is more correct to use disk_is_zero, meaning that reading from the disk will return zeros.

In qemu-img convert this option is called --target-is-zero. The client disk_is_zero try to be similar to qemu-img.

If you don't zero a raw disk you may have random data on the disk. If you create a file system on the disk the data is likley to be hidden, but if you use the disk directly (e.g. database using raw block device) you will have junk data that may break the application.

We also have similar flag in vdsm copy_data.

nirs: The docstring needs a fix, it should be:

    should be used only when uploading to new disk on file based storage or
    new qcow2 disk on block based storage.

aesteve-rh: Ah! I get it. Thanks for the explanations.

And the justification for keeping the name is good too. Even if only for similarity with `qemu-img`, already good reason.

test/client_test.py
aesteve-rh: How is the parameter tested here? Through the qemu-img compare? As in: the resulting disk will be the same as the source image regardless of the disk being zero or not.

ovirt_imageio/client/_upload.py
aesteve-rh: Just an idea, since this new attribute is a derivative of other attributes, I thought it may fit as a `property`. And then the I wondered what other logic could be moved to the class, to justify having it.

Let's keep it in a box for now, for the time being is better to have the logic together, yes.

test/client_test.py
nirs: We test that the content is correct with qemu-img compare.

For shallow upload the value does not matter since we set backing_chain to False. The test verify that setting it to True does not break the upload.

aesteve-rh (approved):

PR #129, opened 2022-08-05
RELEASE.md: small update

aesteve-rh:
Small update to the RELEASE document, that includes:
- Remove the step about downloading the RPMs in Upstream.
  It is not required now.
- Show how to enable the Fedora RCM tools repository for
  installing rhpkg.
- Set public key algorithm for pushing the package with rhpkg.
- Small typo.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -17,13 +17,9 @@
 4. Create a new release in github form the new tag:
    https://github.com/oVirt/ovirt-imageio/releases
 
-5. Download the rpms built by github, and upload them to the new
-   release:
-   https://github.com/oVirt/ovirt-imageio/releases/tag/vX.Y.Z
+5. Publish the release.
 
-6. Publish the release.
-
-7. Send a patch to releng-tools project, adding the new version:
+6. Send a patch to releng-tools project, adding the new version:
    https://github.com/oVirt/releng-tools/commit/97b353b1107ba4bd06776ed4c2e31b309909a357
 
 ### Where to get the packages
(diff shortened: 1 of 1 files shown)


RELEASE.md
```diff
@@ -72,7 +68,20 @@ https://pypi.org/project/ovirt-imageio/
...
+This part also requires having rhpkg installed. To do so, make sure the Fedora
+RCM Tools repository is enabled:
+
+    $ cat /etc/yum.repos.d/rcm-tools-fedora.repo
+    [rcm-tools-fedora-rpms]
+    name=RCM Tools for Fedora $releasever (RPMs)
+    baseurl=https://download.devel.redhat.com/rel-eng/RCMTOOLS/latest-RCMTOOLS-2-F-$releasever/compose/Everything/$basearch/os/
+    enabled=1
+    gpgcheck=1
+    gpgkey=https://download.devel.redhat.com/rel-eng/RCMTOOLS/RPM-GPG-KEY-rcminternal
```
nirs, line 82: Adding this in public document is not a good idea. It will be better to add the
way you can find the "RCM Tools" repos. I think it should be documented in a public web page in the "source" site. The page may be readable only by Red Hat people but the link can be publish in public documentaion.

RELEASE.md
```diff
@@ -117,6 +126,14 @@ This parts works only inside the Red Hat network.
        $ rhpkg push
        $ rhpkg build
 
+   If the push command fails, make sure the correct public key algorithm
+   is used.
+
+       $ cat ~/.ssh/config
+       Host pkgs.devel.redhat.com
+          HostkeyAlgorithms +ssh-rsa
+          PubkeyAcceptedAlgorithms +ssh-rsa
```
nirs, line 135: This is also private information that is not good for public document. You can keep it
if you remove the actual server address, or if you find the public page documenting this
issue. If you don't find a public doc for this issue, we can keep it.

RELEASE.md
```diff
@@ -117,6 +126,14 @@ This parts works only inside the Red Hat network.
        $ rhpkg push
        $ rhpkg build
 
+   If the push command fails, make sure the correct public key algorithm
+   is used.
```
nirs, line 130: This is more correct "enable insecure algorithms used by the old server."

nirs (changes requested):

RELEASE.md
aesteve-rh: I think I found both links :)

nirs (approved):

Issue #125, opened 2022-08-02
Add ovirt-img --version


nirs:
Also log the version when starting in info level to make it easier to collect info for issues.

This can be a global options like --help on the main parser:
```
$ ovirt-img 
usage: ovirt-img [-h] {download-disk,upload-disk} ...

Transfer disk images

options:
  -h, --help            show this help message and exit
  --version             show version and exit

commands:
  {download-disk,upload-disk}
    download-disk       Download disk
    upload-disk         Upload disk
```

PR #130, opened 2022-08-05
ovirt-img: add --version option

aesteve-rh:
Add --version option to ovirt-img tool to show current version:
```
$ ./ovirt-img -h
usage: ovirt-img [-h] [--version] {download-disk,upload-disk} ...

Transfer disk images

optional arguments:
  --version             show program's version number and exit

commands:
  {download-disk,upload-disk}
    download-disk       Download disk
    upload-disk         Upload disk
```
```
$ ./ovirt-img --version
ovirt-img 2.4.6
```
Fixes: #125

### DIFF
--- a/ovirt_imageio/client/_options.py
+++ b/ovirt_imageio/client/_options.py
@@ -13,6 +13,7 @@ Tool options.
 import getpass
 from collections import namedtuple
 
+from .. _internal import version
 from .. _internal.units import KiB, MiB, GiB, TiB
 from . _api import MAX_WORKERS, BUFFER_SIZE
 
@@ -135,6 +136,10 @@ class Parser:
         self._parser = argparse.ArgumentParser(
             description="Transfer disk images")
         self._parser.set_defaults(command=None)
+        self._parser.add_argument(
+            '--version',
+            action='version',
+            version=f'%(prog)s {version.string}')
         self._commands = self._parser.add_subparsers(title="commands")
 
     def add_sub_command(self, name, help, func, transfer_options=True):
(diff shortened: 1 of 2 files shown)


ovirt_imageio/client/_options.py
```diff
@@ -15,6 +15,7 @@
 
 from .. _internal.units import KiB, MiB, GiB, TiB
 from . _api import MAX_WORKERS, BUFFER_SIZE
+from . import __version__
```
nirs, line 18: This can create import cycle (you depend on __init__.py, which may depend on other module which may depend on this module. If you are lucky this works, but when it breaks it is ugly. The only sane way is to always avoid these kinds of imports.

This also use public API when to implement private API. This is always bad idea and 
should be avoided. Never call public API from private API of from another public API. Public API should call only private APIs.

Using:

    from .. _internal import version

is the canonical way to get the version. Please use it like other code accessing the version.

nirs (changes requested):

ovirt_imageio/client/_options.py
aesteve-rh: ACK

nirs (commented):
Looks ready now, but we need to bump version to 2.4.7 first.

nirs (approved):

aesteve-rh:
Since the version matched the release, it didn't register to me as wrong. But of course, this is the next development cycle.

Good that you noticed! I'll take it into account for the next release.

nirs (approved):

PR #131, opened 2022-08-10
Makefile: add venv target

aesteve-rh:
Add venv target to the Makefile, which automatically
creates the Python virtual environment in a
consistent folder and installs all requirements in a
single step.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,8 @@ PACKAGE_NAME=ovirt-imageio
 PACKAGE_VERSION=$(shell python3 ovirt_imageio/_internal/version.py)
 OUTDIR=dist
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -3,6 +3,9 @@ PACKAGE_NAME=ovirt-imageio
 PACKAGE_VERSION=$(shell python3 ovirt_imageio/_internal/version.py)
 OUTDIR=dist
 
+VENV_DIR = $(HOME)/.venv/ovirt-imageio
+VENV = $(VENV_DIR)/bin
```
nirs, line 7: I don't think this is really needed, we used it only twice and it is confusing.

nirs (approved):

Issue #124, opened 2022-08-01
Add tests for ovirt-img sub commands


aesteve-rh:
We meed more tests to increase coverage and avoid breaking on updates for the recent ovirt-img command.

The complexity is that it requires an engine.

One option is add integration tests that contributors need to run manually with engine, or that will run if you enable the target:
```
tox -e integration
```
The test can create an image with `virt-builder`, upload it to configured storage domain, and download it.
We need some kind of local configuration that:
1. enables the integration tests
2. provides the storage domain name

PR #134, opened 2022-08-24
ovirt-img: command testing

aesteve-rh:
Add the required testing infrastructure to test
ovirt-img command with a local setup.

Engine configuration needs to be set in a configuration
file, and then the tests will read this configuration
and run normally through tox.

Fixes: #124


rokkbert:
On my system, with locally running ovirt, the test seems to be too slow to complete within the timeout, when ran for the first time. Maybe my internet connection is not fast enough. Maybe we could use a smaller image?

client run-test: commands[0] | pytest -vs --timeout 200 --log-cli-level=debug test/client
==================================== test session starts ====================================
platform linux -- Python 3.6.8, pytest-7.0.1, pluggy-1.0.0 -- /home/rob/share/git/aesteve-rh/ovirt-imageio2/.tox/client/bin/python
cachedir: .tox/client/.pytest_cache
rootdir: /home/rob/share/git/aesteve-rh/ovirt-imageio2, configfile: tox.ini
plugins: timeout-2.1.0
timeout: 200.0s
timeout method: thread
timeout func_only: False
collected 2 items                                                                           

test/client/upload_download_test.py::test_upload_download[raw] [   0.9] Downloading: http://builder.libguestfs.org/fedora-36.xz
###################################################################################### 100.0%########################################################                                65.4%
++++++++++++++++++++++++++++++++++++++++++ Timeout ++++++++++++++++++++++++++++++++++++++++++

~~~~~~~~~~~~~~~~~~~~~~~~~~~ Stack of MainThread (140299774200704) ~~~~~~~~~~~~~~~~~~~~~~~~~~~
########################################################                                65.4%  File "/home/rob/share/git/aesteve-rh/ovirt-imageio2/.tox/client/bin/pytest", line 8, in <module>
    sys.exit(console_main())
  File "/home/rob/share/git/aesteve-rh/ovirt-imageio2/.tox/client/lib/python3.6/site-packages/_pytest/config/__init__.py", line 188, in console_main
    code = main()

rokkbert:
Also, the test freezes for me, maybe because I don't have enough space to upload the image:

                   Output file: /var/tmp/imageio-tests/test_upload_download_raw_0/image.raw
                   Output size: 6.0G
                 Output format: raw
            Total usable space: 6.0G
                    Free space: 4.7G (79%)

-------------------------------------------------------------------------------- live log call ---------------------------------------------------------------------------------
15:02:42,352 INFO    (MainThread) [test] Upload raw image to SD nfs1
[ ---- ] 0 bytes, 0.01 s, 0 bytes/s | creating disk

aesteve-rh:
@rokkbert thank you for trying the test!

The initial timeout is due to `virt-builder`, probably I had the image already cached and is faster, I didn't think that the first time it will take a while to download. The disk could be smaller, but the idea was to try a realistic disk with data. I could increase the timeout for the tox env, or maybe it would be better to use `qemu-img` and write a few bytes into it before uploading, that will be a lot faster. I'll think about it.

About the freeze, no sure the reason. How much free space do you have in the nfs1 SD? Does it consistently freeze in the same step?

rokkbert:
> The initial timeout is due to `virt-builder`, probably I had the image already cached and is faster, I didn't think that the first time it will take a while to download.

Yes, the second time it is fast. Maybe not worth the effort of finding a workaround, but maybe the image could be downloaded in advance, before the test and its timeout become active.

> The disk could be smaller, but the idea was to try a realistic disk with data.

Makes sense.

> I could increase the timeout for the tox env, or maybe it would be better to use `qemu-img` and write a few bytes into it before uploading, that will be a lot faster. I'll think about it.

I think 400-500 seconds would be enough for my case, but internet in germany isn't that fast, so maybe no an issue for others. ;-)

> About the freeze, no sure the reason. 

Reason was that I made a mistake in the configuration :)

But now I'm out of space on my test-VM, which is a bit more work to fix.

rokkbert:
After changing the tmp-path in tox.ini to somewhere with more space the tests run successfully.

I just get 2 deprecation warnings, everything else is fine:

.tox/client/lib/python3.6/site-packages/_pytest/config/__init__.py:1196
  /home/rob/share/git/aesteve-rh/ovirt-imageio2/.tox/client/lib/python3.6/site-packages/_pytest/config/__init__.py:1196: PytestRemovedIn8Warning: The --strict option is deprecated, use --strict-markers instead.
    _pytest.deprecated.STRICT_OPTION, stacklevel=2

.tox/client/lib/python3.6/site-packages/userstorage/config.py:21
  /home/rob/share/git/aesteve-rh/ovirt-imageio2/.tox/client/lib/python3.6/site-packages/userstorage/config.py:21: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
    import imp

aesteve-rh:
`But now I'm out of space on my test-VM, which is a bit more work to fix.` That's too bad! It can be tested from your local development machine (at least I do test from there). As long as the engine configuration is correct, it should be fine.

`After changing the tmp-path in tox.ini to somewhere with more space the tests run successfully.` Which path do you mean?

Warnings are "ok", is not related to this PR. The pytest flags in `tox.ini` need an update.

Again, thanks for testing!

rokkbert:
I mean this path, basetemp, in tox.ini:
addopts = -rsxX --basetemp=/var/tmp/imageio-tests

rokkbert (approved):
It works for me and looks fine.

PR #135, opened 2022-08-24
http: avoid double slashes in url

aesteve-rh:
Recent HTTPRequestHandler changes avoid urls that
start with '//' to protect against open redirect
attacks (see https://github.com/python/cpython/commit/4abab6b603dd38bec1168e9a37c40a48ec89508e).

Server url in the http test have no path,
therefore the handler that is used in the
test receives the root path, resulting in
requests starting with '//' when the Backend
tries to get the extents, which break in the
pipelines for recent Python 3.10 versions.

This is not a realistic scenario and not
something the library needs to handle.
Is better to change the url to a more
realistic scenario, and since we want
to test image handling, have
`/images/ticket-id` in the url path.

### DIFF
--- a/test/backends/http_test.py
+++ b/test/backends/http_test.py
@@ -38,7 +38,7 @@ def http_server(tmp_pki):
     server.socket = ctx.wrap_socket(server.socket, server_side=True)
 
     server.url = urlparse(
-        "https://localhost:{}/".format(server.server_port))
+        "https://localhost:{}/images/ticket-id".format(server.server_port))
     server.cafile = tmp_pki.cafile
     server.app = http.Router([])
 
@@ -79,7 +79,7 @@ class Handler:
         self.dirty = False
         self.requests = 0
 
-        router = http.Router([("/(.*)", self)])
+        router = http.Router([(r"/images/(.*)", self)])
         http_server.app = router
         if uhttp_server:
             uhttp_server.app = router
@@ -197,7 +197,7 @@ class Daemon(Handler):
         """
         Override to dispatch "GET /extents" resource.
         """
-        if path == "/extents":
+        if path == "ticket-id/extents":
             self.requests += 1
             context = req.query.get("context", "zero")
             self._extents(resp, context)


ovirt_imageio/_internal/backends/http.py
```diff
@@ -504,7 +505,8 @@ def _options(self):
         return options
 
     def _get_extents(self, context):
-        self._con.request("GET", self.url.path + "/extents?context=" + context)
```
nirs, line 507: Are we testing extents with path=/?

This is wrong, the application serve only /images/ticket-id/extents.

Better fix the test to match the real code. This is not a library that need to handle any crazy input.

Regardless, joining without creating double // is good change. But the fact that you need to convert the joined path to string is worrying for code in the data path.

aesteve-rh: > Are we testing extents with path=/?

Yes. And then the `Router` regex and the test `Handler` accounted for the extra `/`. I'll update the test as you suggest.

> Regardless, joining without creating double // is good change. But the fact that you need to convert the joined path to  string is worrying for code in the data path.

I agree in that it is a good change. I thought in doing the fix only in the test (since in a real scenario this should not happen), but in the end using `urljoin` sounds correct. Other tests use a more realistic url with `/images/ticket-id/extents`, so passing the pipelines should be safe enough.

test/backends/http_test.py
```diff
@@ -38,7 +38,7 @@ def http_server(tmp_pki):
     server.socket = ctx.wrap_socket(server.socket, server_side=True)
 
     server.url = urlparse(
-        "https://localhost:{}/".format(server.server_port))
+        "https://localhost:{}/images/ticket-id/".format(server.server_port))
```
nirs, line 41: Our url never ends with a slash - it will never match our routes.

test/backends/http_test.py
```diff
@@ -79,7 +79,7 @@ def __init__(self, http_server, uhttp_server=None):
         self.dirty = False
         self.requests = 0
 
-        router = http.Router([("/(.*)", self)])
+        router = http.Router([(r"/images/.*/(.*)", self)])
```
nirs, line 82: The real routes are:

        self._server.app = http.Router([
            (r"/images/(.*)/extents", extents.Handler(config, auth)),
            (r"/images/(.*)/checksum/algorithms",
                checksum.Algorithms(config, auth)),
            (r"/images/(.*)/checksum", checksum.Checksum(config, auth)),
            (r"/images/(.*)", images.Handler(config, auth)),
            (r"/info/", info.Handler(config, auth)),
        ])

This is not possible with the current bad way extents is integrated in this
fake handler, since we use one handler, while the real code uses separate
handler for /images/(.*)/extents.

But we can make this simpler using:

    r"/images/(.*)"

and check later:

   if path == "ticket-id/extents":

aesteve-rh: Ok. Different flavors of the same solution.

bennyz (dismissed):

barpavel:
> Recent HTTPRequestHandler changes avoid urls that start with '//' to protect against open redirect attacks (see [python/cpython@4abab6b](https://github.com/python/cpython/commit/4abab6b603dd38bec1168e9a37c40a48ec89508e)).
> 
> Server url in the http test have no path, therefore the handler that is used in the test receives the root path, resulting in requests starting with '//' when the Backend tries to get the extents, which break in the pipelines for recent Python 3.10 versions.
> 
> This is not a realistic scenario and not something the library needs to handle. Is better to change the url to a more realistic scenario, and since we want to test image handling, have `/images/ticker-id` in the url path.
> 
> Signed-off-by: Albert Esteve [aesteve@redhat.com](mailto:aesteve@redhat.com)

`ticker-id` --> `ticket-id` :)

barpavel (approved):

nirs (approved):

Issue #117, opened 2022-07-29
Handle disk status errors while waiting for disk creation


aesteve-rh:
ovirt-img probes a disk after creation during a fixed time, until it reaches an OK status.

If there is an error in the creation and disk becomes `DiskStatus.ILLEGAL`, users have to wait for the timeout.

We need to handle these error status and fail loudly.

It can also be interesting to allow users to configure a custom timeout for disk creation (through config file probably best).

PR #137, opened 2022-08-26
ovirt-img: disk creation loop

aesteve-rh:
On disk-upload, ovirt-img  probes a disk after
creation during a fixed time, until it reaches an
OK status (or timeout).

If there is an error in the creation and disk becomes
DiskStatus.ILLEGAL, users have to wait for the timeout.

We need to handle this error status and fail loudly.

Add an option, only available through config file,
to change the timeout waiting for disk creation.

This is a fine-gran configuration that should be
fine as default for most cases, but may be
interesting to modify it for stress tests,
or slow networks, etc.

Fixes: https://github.com/oVirt/ovirt-imageio/issues/117

### DIFF
--- a/ovirt_imageio/client/_options.py
+++ b/ovirt_imageio/client/_options.py
@@ -24,6 +24,9 @@ import sys
 import uuid
 
 
+ADD_DISK_TIMEOUT = 120
+
+
 class Choices:
 
     def __init__(self, name, values):
@@ -115,6 +118,12 @@ class Parser:
             config=True,
             help=("Path to ovirt-engine CA certificate. If not set, read "
                   "from the specified config section"),
+        ),
+        Option(
+            name="disk_timeout",
+            config=True,
+            type=int,
+            default=ADD_DISK_TIMEOUT,
         ),
         Option(
             name="log_file",
--- a/ovirt_imageio/client/_ovirt.py
+++ b/ovirt_imageio/client/_ovirt.py
@@ -16,6 +16,8 @@ import time
 import ovirtsdk4 as sdk
 from ovirtsdk4 import types
 
+from . _options import ADD_DISK_TIMEOUT
+
 log = logging.getLogger("ovirt")
 
 # Image transfer constants.
@@ -61,7 +63,7 @@ def find_disk(con, disk_id):
 
 def add_disk(con, name, provisioned_size, sd_name, id=None,
              initial_size=None, sparse=True, enable_backup=True,
-             content_type=DATA, format=COW):
+             content_type=DATA, format=COW, timeout=ADD_DISK_TIMEOUT):
     """
     Add a new disk to the storage domain, based on the source image
     information provided.
@@ -79,6 +81,8 @@ def add_disk(con, name, provisioned_size, sd_name, id=None,
         content_type (ovirtsdk4.types.DiskContentType): Content type for the
             new disk.
         format (ovirtsdk4.types.DiskFormat): Format of the new disk.
+        timeout (int, optional): number of seconds to wait for
+            disk to be created.
 
     Returns:
         ovirtsdk4.types.Disk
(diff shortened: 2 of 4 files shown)


ovirt_imageio/client/_ovirt.py
```diff
@@ -425,6 +428,9 @@ def _wait_for_disk(con, disk_id):
         if time.monotonic() > deadline:
             raise RuntimeError(f"Timeout reached waiting for disk {disk_id}")
 
+        if disk_service.get().status == types.DiskStatus.ILLEGAL:
+            raise RuntimeError(f"Disk {disk_id} failed")
+
         if disk_service.get().status == types.DiskStatus.OK:
```
rokkbert, line 434: Probably not important, but maybe we save the status once, and then compare it, instead of getting it multiple times, once for each comparison?

ovirt_imageio/client/_ovirt.py
```diff
@@ -425,6 +428,9 @@ def _wait_for_disk(con, disk_id):
         if time.monotonic() > deadline:
             raise RuntimeError(f"Timeout reached waiting for disk {disk_id}")
 
+        if disk_service.get().status == types.DiskStatus.ILLEGAL:
```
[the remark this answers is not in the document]
aesteve-rh: `DiskStatus.LOCKED can not happen in this situation, I assume?` If I'm not mistaken, the disk starts as `LOCKED`, until the add disk operation finishes, which updates its status to `OK`.

`What can cause us to exceed the timeout then? Just an unresponsive engine?` Could be also a slow network, I guess? The timeout is quite generous, so I assume it won't happen unless the engine becomes unresponsive mid-operation, or heavy network outages. Everything else should cause the disk to go to `ILLEGAL`.

ovirt_imageio/client/_ovirt.py
aesteve-rh: Yeah I thought about it, but in the end it was one more line with an extra variable and decided otherwise. The reasoning was that this looked slightly better, and the performance is not that important in a wait loop.

But now that you've mentioned, I looked at it again, and I think you are right, so I probably change it lol

rokkbert: I guess the extra requests matter more for the engine than the client.

ovirt_imageio/client/_options.py
```diff
@@ -24,6 +24,9 @@
 import uuid
 
 
+DISK_TIMEOUT = 120
```
rokkbert, line 27: If the timeout is only used for waiting until the engine has created the new disk then maybe we could give it a more descriptive name, like DISK_CREATION_TIMEOUT or something like that, because in the general context of "DISK" there could be different timeouts of other operations as well.

rokkbert (approved):

ovirt_imageio/client/_ovirt.py
```diff
@@ -425,7 +425,11 @@ def _wait_for_disk(con, disk_id):
         if time.monotonic() > deadline:
             raise RuntimeError(f"Timeout reached waiting for disk {disk_id}")
 
-        if disk_service.get().status == types.DiskStatus.OK:
+        disk = disk_service.get()
+        if disk.status == types.DiskStatus.ILLEGAL:
+            raise RuntimeError(f"Disk {disk_id} failed")
```
nirs, line 433: The error message hides the real status of the disk. It will be more useful if the error includes the disk id and the status of the disk:

    Disk xxx-yyy became ILLEGAL

ovirt_imageio/client/_options.py
```diff
@@ -24,6 +24,9 @@
 import uuid
 
 
+DISK_CREATION_TIMEOUT = 120
```
nirs, line 27: ADD_DISK_TIMEOUT? (since we call the operation add_disk)

nirs (changes requested):

nirs (approved):

PR #138, opened 2022-09-12
ci: build containers automatically every week

aesteve-rh:
Build project containers (currently four)
and push them every week to update at least
the underlaying OS.

Fixes: #13

### DIFF
--- a/.github/workflows/containers.yml
+++ b/.github/workflows/containers.yml
@@ -39,3 +39,25 @@ jobs:
       - name: Print image url
         if: ${{ env.can_push == 'true' }}
         run: echo "Image pushed to ${{ steps.push-to-quay.outputs.registry-paths }}"
+  container:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install dependencies
+        run: |
+          sudo apt update
+          sudo apt install podman
+      - name: Build container images
+        run: make container
+      - name: Push to Quay.io
+        if: ${{ env.can_push == 'true' }}
+        id: push-to-quay
+        uses: redhat-actions/push-to-registry@v2
+        with:
+          image: ovirt/ovirt-imageio
+          registry: ${{ env.IMAGE_REGISTRY }}
+          username: ${{ secrets.QUAY_USERNAME  }}
+          password: ${{ secrets.QUAY_TOKEN }}
+      - name: Print image url
+        if: ${{ env.can_push == 'true' }}
+        run: echo "Image pushed to ${{ steps.push-to-quay.outputs.registry-paths }}"
--- a/.github/workflows/containers.yml
+++ b/.github/workflows/containers.yml
@@ -1,0 +1,41 @@
+# SPDX-FileCopyrightText: Red Hat, Inc.
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+name: Containers
+on:
+  # Be able to run the job manually when needed
+  # (Actions -> Containers -> Run workflow)
+  workflow_dispatch:
+  # Build every week on Monday 00:00
+  schedule:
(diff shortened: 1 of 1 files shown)


aesteve-rh:
I do not have permission to add secrets. Maybe you can, @ahadas ?

Once the branch is merged, we can add the secrets and test the action manually.

.github/workflows/containers.yml
```diff
@@ -0,0 +1,38 @@
...
+      matrix:
+        container: [ centos-8 centos-9 fedora-35 fedora-36 ]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install dependencies
+        run: |
+          sudo apt update
+          sudo apt install podman
+      - name: Build container images
+        working-directory: automation/containers
```
nirs, line 25: containers?

.github/workflows/containers.yml
```diff
@@ -0,0 +1,38 @@
...
+        run: make ${{ matrix.container }}
+      - name: Push to Quay.io
+        id: push-to-quay
```
nirs, line 28: Do we really need an id?

.github/workflows/containers.yml
```diff
@@ -0,0 +1,38 @@
...
+        uses: redhat-actions/push-to-registry@v2
+        with:
+          image: ovirt/${{ matrix.container }}
+          tags: ${{ matrix.container }}
```
nirs, line 32: I don't think we have a tag, are you sure this is correct?

.github/workflows/containers.yml
```diff
@@ -0,0 +1,38 @@
...
+          registry: ${{ env.IMAGE_REGISTRY }}
+          username: ${{ secrets.QUAY_USERNAME  }}
+          password: ${{ secrets.QUAY_TOKEN }}
+      - name: Print image url
+        if: ${{ env.can_push == 'true' }}
+        run: echo "Image pushed to ${{ steps.push-to-quay.outputs.registry-paths }}"
```
nirs, line 38: So we try to push even when we cannot push?

I expect that we use redhat-actions/push-to-registry@v2 only if we can push.

.github/workflows/containers.yml
aesteve-rh: argh, true.

.github/workflows/containers.yml
aesteve-rh: Yeah, only tags for imageio are the default repository ones. If we don't use tags I can remove this part.

.github/workflows/containers.yml
aesteve-rh: Yes, my bad, missed that line.

.github/workflows/containers.yml
aesteve-rh: It is used below in the next step: `steps.push-to-quay.outputs.registry-paths` , and is consistent with similar workflows in other ovirt projects. It does not harm.

.github/workflows/containers.yml
```diff
@@ -0,0 +1,41 @@
...
+    - cron:  '0 0 * * 1'
+env:
+  IMAGE_REGISTRY: quay.io
+  can_push: ${{ github.repository_owner == 'oVirt' }}
+jobs:
+  test-containers:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        container: [ centos-8 centos-9 fedora-35 fedora-36 ]
```
nirs, line 20: distro?

nirs (approved):
Looks good, we can tweak the matrix later.

.github/workflows/containers.yml
aesteve-rh: I'll prepare a quick follow up branch for this, plus whatever other tweak may be required after we test the workflow,

PR #139, opened 2022-09-20
update userstorage version

aesteve-rh:
Update userstorage to v0.5.1.

Set pytest.xfail as handler for missing
storage.

Currently, all tests that rely on userstorage employ
the Backend wrapper only to add a couple attributes.

With recent update, context management is natively
supported by userstorage library. On the other
hand, not all test require the Backend additional
attributes.

We can avoid using the Backend wrapper completely
for tests that do not require the attributes.
Otherwise, the Backend can just wrap the
storage from within the userstorage context (i.e.,
after the storage is setup), so we avoid the
__enter__() and __exit__() overrides.

### DIFF
--- a/test/backends/file_test.py
+++ b/test/backends/file_test.py
@@ -44,8 +44,9 @@ KERNEL_VERSION = tuple(map(int, os.uname()[2].split(".")[:2]))
     ids=lambda x: str(x[0]),
 )
 def user_file(request):
-    with storage.Backend(*request.param) as backend:
-        yield backend
+    backend, can_detect_sector_size = request.param
+    with backend:
+        yield storage.Backend(backend, can_detect_sector_size)
 
 
 def test_debugging_interface(user_file):
--- a/test/backends/nbd_test.py
+++ b/test/backends/nbd_test.py
@@ -33,8 +33,9 @@ BACKENDS = userstorage.load_config("storage.py").BACKENDS
     ids=str
 )
 def user_file(request):
-    with storage.Backend(request.param) as backend:
-        yield backend
+    backend = request.param
+    with backend:
+        yield storage.Backend(backend)
 
 
 def test_debugging_interface(nbd_server):
--- a/test/handlers/checksum_test.py
+++ b/test/handlers/checksum_test.py
@@ -21,7 +21,6 @@ from ovirt_imageio._internal import server
 
 from .. import testutil
 from .. import http
-from .. import storage
 
 BACKENDS = userstorage.load_config("storage.py").BACKENDS
 ALGORITHMS = frozenset(hashlib.algorithms_available)
@@ -35,7 +34,8 @@ ALGORITHMS = frozenset(hashlib.algorithms_available)
     ids=str
 )
 def user_file(request):
-    with storage.Backend(request.param) as backend:
+    backend = request.param
+    with backend:
         yield backend
 
 
--- a/test/storage.py
+++ b/test/storage.py
@@ -8,28 +8,29 @@
 
 import urllib.parse
 
-import pytest
-
 
 class Backend:
     """
-    Wrap a userstorage.Backend, adding a url and context manager interface to
-    simplify fixtures.
+    Wrap a userstorage.Backend, adding a url and the can_detect_sector_size
+    attribute.
     """
 
     def __init__(self, storage, can_detect_sector_size=True):
-        if not storage.exists():
-            pytest.xfail("Storage {} is not available".format(storage.name))
-
         self._storage = storage
-        self.path = storage.path
         self.url = urllib.parse.urlparse("file:" + storage.path)
-        self.sector_size = storage.sector_size
         self.can_detect_sector_size = can_detect_sector_size
 
+    @property
+    def path(self):
+        return self._storage.path
+
+    @property
+    def sector_size(self):
+        return self._storage.sector_size
+
     def __enter__(self):
-        self._storage.setup()
+        self._storage.__enter__()
         return self
 
     def __exit__(self, *args):
-        self._storage.teardown()
+        self._storage.__exit__(*args)
--- a/test/storage.py
+++ b/test/storage.py
@@ -27,10 +27,3 @@ class Backend:
     @property
     def sector_size(self):
         return self._storage.sector_size
-
-    def __enter__(self):
-        self._storage.__enter__()
-        return self
-
-    def __exit__(self, *args):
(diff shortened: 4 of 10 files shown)


test/storage.py
```diff
@@ -9,27 +9,34 @@
...
+        return self._storage.sector_size
+
     def __enter__(self):
-        self._storage.setup()
+        self._storage.__enter__()
         return self
 
     def __exit__(self, *args):
-        self._storage.teardown()
+        self._storage.__exit__(args)
```
nirs, line 42: Must be:

    self._storage.__exit__(*args)

But it will be better to use the standard signature[1]:

    def __exit__(self, t, v, tb):
        self._storage.__exit__(t, v, tb)

[1] https://docs.python.org/3/reference/datamodel.html#object.__exit__

nirs (changes requested):

test/storage.py
```diff
@@ -9,27 +9,34 @@
 import urllib.parse
 
 import pytest
+import userstorage
+
+
+userstorage.missing_handler = pytest.xfail
```
nirs, line 15: Lets move this to conftest.py?

test/storage.py
aesteve-rh: I agree. But I'll just use it the `self._storage.__exit__(*args)` version, since I just remove the context management from the `Backend` class in the next commit (after what we discussed).

test/backends/file_test.py
```diff
@@ -44,8 +44,9 @@
     ids=lambda x: str(x[0]),
 )
 def user_file(request):
-    with storage.Backend(*request.param) as backend:
-        yield backend
+    backend_file, can_detect_sector_size = request.param
+    with backend_file as backend:
+        yield storage.Backend(backend, can_detect_sector_size)
```
nirs, line 49: We don't really need to get the result of the context manager, so we can do:

    backend, can_detect_sector_size = request.param
    with backend:
        yield storage.Backend(backend, can_detect_sector_size)

test/backends/nbd_test.py
```diff
@@ -33,8 +33,8 @@
     ids=str
 )
 def user_file(request):
-    with storage.Backend(request.param) as backend:
-        yield backend
+    with request.param as backend:
+        yield storage.Backend(backend)
```
nirs, line 37: we don't need to wrap the backend here, it can be handled like checksum_test.py

test/handlers/checksum_test.py
```diff
@@ -35,7 +34,7 @@
     ids=str
 )
 def user_file(request):
-    with storage.Backend(request.param) as backend:
+    with request.param as backend:
         yield backend
```
nirs, line 38: Since `userstorage.Base.__enter__()` returns self, we don't really need to get the
return value of the context manager, so we can simply do:

    with request.param:
        yield request.param

nirs (changes requested):

test/handlers/checksum_test.py
aesteve-rh: This is technically the same as current code? Just different ways of aliasing `request.param`, no?
```
backend = request.param
with backend:
    yield backend
```

nirs: Current code yields the the result of the with statement, which is the same since we backend.__enter__() returns self.

Both options work, my suggestion was to use the simplest possible form.

aesteve-rh: I understand why they result in the same behavior, but I do not really see how it is simpler or better assigning the parameter before the with statement or using the result of the with-as statement. Different flavors of the same code.

Using the result of the with statement has the advantage of:
- with-as is a recognizable statement
- if in the future userstorage changes return in `__enter__` (unlikely...), we get the new returned object already by updating
- is already coded like that in the current version!

nirs: The current code is confusing:

    with request.param as backend:

This looks like request.param is not a backend and the result of the with statement is a backend. This confuses the reader that there is something going on under the hood while there is no such thing.

The suggested code:

    with request.param:
        yield request.param

I clear that we must yield request.param in the context of a with statement. But this does not reveal what is request.param, this is why I suggest the option to have an explaining variable:

     backend = request.param

Just to make it clear that the fixture is accepting a backend as parameter.

There is also another issue caused by using the result of the with statement. In the case when we request.param is a tuple, we want to unpack it:

    backend, can_detect_sector_size = request.param

In the current code you invented a backend_file (is this the file of a backend?) so we can use the name backend for the result of the with statement. If we don't use the result of the with statement we don't need to invent new names.

Usually context manager return self. In some cases like tempfile.TemporaryDirectory this is not the case and the result of the with statement is a string instead of the temporary directory object. This is unexpected and confusing.

Future version of userstorage cannot return different value from the with statement since this will be backward incompatible with current users of the code not using the result of the with statement.

aesteve-rh: I understand the change for the cases that require the tuple, I will do the change.

For this case it feels like saying that this is confusing:
```
with open('file', 'r') as f:
     f.read()
```
And is clearer to do:
```
f = open('file', 'r')
with f:
    f.read()
```
I can see what got you there with what you explained. But not sure, I do not see it confusing. Maybe a little obscure as you can only assume what the `__enter__` is returning, but the same happens in practice every time we use a context. In general, you just expect it to behave.

But anyway, if you really think this is clearer, I'll do the change tomorrow.

nirs: The open example is not the same. Calling open() returns an open file object that can be used as context manager to ensure the file is closed. Creating a temporary variable is typically a bad idea since it can be broken by adding another line after creating the file. If the other line raises, we will not close the open file. `f` is the common name for an open file, so everything is clear.

In our case we have a request.param, which is already a backend, but this is not clear when you read the code. We need to use a with statement with the backend and yield it to the caller - nothing else is needed. The suggested code does exactly this.

This is little detail, but we are going to repeat this code in 2 projects, so I think it worth the time to do this change in the best way.

aesteve-rh: True, both are already a backend, but one of them is `setup()`, and will be `teardown()` at the end. We could maybe use `backend`, or `b` to signify that a temporary variable is a setup backend.

nirs: I don't think we need a different name for a setup backend. Inside the context, the backend is set up, and outside the context it is not.

nirs (approved):

PR #140, opened 2022-09-22
Load ticket as the Server starts

aesteve-rh:
Add a -t/--ticket option to the ovirt-imageio
server CLI, so that it can optionally load a
ticket directly as the server starts.
```
$ ovirt-imageio -c config -t path/to/ticket.json
2022-09-22 12:36:51,082 INFO    (MainThread) [server] Starting (hostname=fedora pid=3113040, version=2.4.7)                                                   
2022-09-22 12:36:51,082 DEBUG   (MainThread) [services] Creating remote.service on port 54322                                                                  
...
2022-09-22 12:36:51,087 INFO    (MainThread) [server] Initial ticket: examples/file.json
2022-09-22 12:36:51,087 INFO    (MainThread) [services] control.service listening on 'test/daemon.sock'
2022-09-22 12:36:51,087 DEBUG   (MainThread) [services] Starting remote.service 
2022-09-22 12:36:51,087 DEBUG   (remote.service) [services] remote.service started
2022-09-22 12:36:51,087 DEBUG   (MainThread) [services] Starting local.service
2022-09-22 12:36:51,087 DEBUG   (local.service) [services] local.service started
2022-09-22 12:36:51,087 DEBUG   (MainThread) [services] Starting control.service
2022-09-22 12:36:51,087 DEBUG   (control.service) [services] control.service started
2022-09-22 12:36:51,088 INFO    (MainThread) [server] Ready for requests
```

### DIFF
--- a/ovirt_imageio/_internal/server.py
+++ b/ovirt_imageio/_internal/server.py
@@ -23,6 +23,7 @@ import systemd.daemon
 
 from . import auth
 from . import config
+from . import errors
 from . import services
 from . import version
 
@@ -44,7 +45,7 @@ def main():
         log.info("Starting (hostname=%s pid=%s, version=%s)",
                  socket.gethostname(), os.getpid(), version.string)
 
-        server = Server(cfg)
+        server = Server(cfg, ticket=args.ticket)
         signal.signal(signal.SIGINT, server.terminate)
         signal.signal(signal.SIGTERM, server.terminate)
 
@@ -75,6 +76,9 @@ def parse_args():
         help="print actual configuration in json format and exit. This is "
              "useful for debugging configuration issues, or reading imageio "
              "configuration by other programs.")
+    parser.add_argument(
+        "-t", "--ticket",
+        help="path to a ticket to load during server startup.")
     return parser.parse_args()
 
 
@@ -114,7 +118,7 @@ def configure_logger(cfg):
 
 class Server:
 
-    def __init__(self, config):
+    def __init__(self, config, ticket=None):
         self.config = config
         self.running = False
         self.auth = auth.Authorizer(config)
@@ -126,6 +130,9 @@ class Server:
 
         if os.geteuid() == 0 and self.config.daemon.drop_privileges:
             self._drop_privileges()
+
+        if ticket:
+            self._add_ticket(ticket)
 
     def start(self):
         assert not self.running
@@ -177,3 +184,17 @@ class Server:
         os.initgroups(self.config.daemon.user_name, gid)
         os.setgid(gid)
         os.setuid(uid)
+
+    def _add_ticket(self, ticket):
+        log.info("Initial ticket: %s", ticket)
+        self.auth.add(self._read_ticket(ticket))
+
+    def _read_ticket(self, ticket):
+        try:
+            with open(ticket, 'r') as f:
+                return json.loads(f.read())
+        except ValueError as e:
+            raise errors.InvalidTicket(
+                f"Cannot parse ticket {ticket}: {e}") from e
+        except FileNotFoundError as e:
+            raise errors.InvalidTicket(f"Cannot read ticket: {e}") from e
--- a/test/containerized_test.py
+++ b/test/containerized_test.py
@@ -1,0 +1,46 @@
+# Copyright (C) 2022 Red Hat, Inc.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+
+import json
+import http.client as http_client
+
+from ovirt_imageio._internal import config
+from ovirt_imageio._internal import server
+from ovirt_imageio._internal.units import KiB
+
+from . import testutil
+from . import http
+
+
+TICKET_SIZE = 16 * KiB
+
+
+def test_start_with_ticket(tmpdir):
+    # Create ticket
+    image = testutil.create_tempfile(tmpdir, name="disk.raw", size=TICKET_SIZE)
+    ticket = testutil.create_ticket(size=TICKET_SIZE, url=f'file://{image}')
+    ticket_id = ticket['uuid']
+    ticket_path = tmpdir.join("file.json")
+    ticket_path.write(json.dumps(ticket))
+
+    # Start server with ticket
+    cfg = config.load("test/conf/daemon.conf")
+    srv = server.Server(cfg, ticket=ticket_path)
+    srv.start()
+    try:
+        data = b"a" * (TICKET_SIZE // 2) + b"b" * (TICKET_SIZE // 2)
+        # Test that we can upload
+        with http.RemoteClient(srv.config) as c:
+            res = c.put(f'/images/{ticket_id}', data)
+            assert res.status == http_client.OK
+        # Test that we can download and matches the uploaded data
+        with http.RemoteClient(srv.config) as c:
+            res = c.get(f'/images/{ticket_id}')
+            assert res.read() == data
+            assert res.status == http_client.OK
+    finally:
+        srv.stop()


ovirt_imageio/_internal/handlers/tickets.py
```diff
@@ -21,9 +21,19 @@ class Handler:
     Handle requests for the /tickets/ resource.
     """
 
-    def __init__(self, config, auth):
+    def __init__(self, config, auth, ticket=None):
```
nirs, line 24: A handler is never created with a ticket. We don't want to add such argument and have inconsistent signature.

We can handle this in the server module - accept a --ticket argument, and add the ticket to the autorizer. The request handler does not need to know anything about initial ticket.

All the changes in this module are not needed.

test/server_test.py
```diff
@@ -39,11 +39,11 @@
 port = 0
 
 [local]
-socket =
+socket = {run_dir}/local.sock
```
nirs, line 44: The local socket is never unix socket, it is always an absract unix socket - using file path is incorrect since this is not the way the socket is used.

The user or the tests never need to know the path of the socket, we report the socket in OPTIONS request.

test/server_test.py
```diff
@@ -39,11 +39,11 @@
...
 
 [control]
 transport = unix
-socket = {run_dir}/sock
+socket = {run_dir}/control.sock
```
nirs, line 48: This is good change regardless of using a local socket, since it is more clear that the socket is the control socket.

But this change is not related to support initial ticket, please split it to another pr to make this pr easier to review.

test/server_test.py
```diff
@@ -226,7 +227,7 @@ def started_imageio(tmpdir, drop_privileges="true"):
             with http.ControlClient(cfg) as c:
                 r = c.get("/tickets/no-such-ticket")
                 r.read()
-                assert r.status == 404
+                assert r.status == http_client.NOT_FOUND
```
nirs, line 249: This change is not related to supporting initial ticket, please split to another pr.

ovirt_imageio/_internal/server.py
```diff
@@ -114,15 +117,16 @@ def configure_logger(cfg):
...
         self.config = config
         self.running = False
         self.auth = auth.Authorizer(config)
         self.remote_service = services.RemoteService(self.config, self.auth)
         self.local_service = None
         if config.local.enable:
             self.local_service = services.LocalService(self.config, self.auth)
-        self.control_service = services.ControlService(self.config, self.auth)
+        self.control_service = services.ControlService(
+            self.config, self.auth, ticket)
```
nirs, line 129: No need to modify the control service - we initialize the service with the auth instance. The only thing we need to do at this point is to add the ticket to the auth:

    self.auth.add(read_ticket(ticket))

And no change is needed in the control service - which can and should be disable when using containerize deployment!

ovirt_imageio/_internal/services.py
```diff
@@ -145,7 +145,7 @@ class ControlService(Service):
 
     name = "control.service"
 
-    def __init__(self, config, auth):
+    def __init__(self, config, auth, ticket=None):
```
nirs, line 148: Changes in this module are not needed

test/server_test.py
```diff
@@ -65,6 +66,22 @@ def tmp_dirs(tmpdir):
     tmpdir.mkdir("conf").mkdir("conf.d")
 
 
+@pytest.fixture
+def tmp_image(tmpdir):
+    return testutil.create_tempfile(tmpdir, "disk.raw")
+
+
+@pytest.fixture
+def ticket(tmp_image, tmpdir):
```
nirs, line 75: We use tmp_ prefix for such fixtures.

test/server_test.py
```diff
@@ -222,7 +241,7 @@ def started_imageio(tmpdir, drop_privileges="true"):
 
         # Wait until server is listening - at this point it already dropped
         # privileges.
-        if drop_privileges:
+        if drop_privileges and ticket is None:
```
nirs, line 244: why dropping privileges is related to not having a ticket? These features should not have any dependencies.

test/server_test.py
```diff
@@ -235,6 +254,23 @@ def started_imageio(tmpdir, drop_privileges="true"):
         proc.wait()
 
 
+@requires_root
+def test_load_ticket(tmpdir, tmp_dirs, tmp_image, ticket):
```
nirs, line 258: test_start_with_ticket? test_initial_ticket?

test/server_test.py
```diff
@@ -235,6 +254,23 @@ def started_imageio(tmpdir, drop_privileges="true"):
...
+    data = b"a" * 512 + b"b" * 512
+    tmp_image.write(data)
+    with started_imageio(tmpdir, drop_privileges="false", ticket=ticket):
+        conf_dir = tmpdir.join("conf")
+        cfg = config.load(str(conf_dir.join("conf.d", "daemon.conf")))
+        with http.ControlClient(cfg) as c:
+            res = c.get('/tickets/file')
+            assert res.status == http_client.OK
```
nirs, line 266: We don't need the control service to check that we have a ticket. Do upload/download
using this ticket instead.

The typical configuration when using initial ticket is to disable the control service since in this mode we serve exactly one ticket - so we don't have a control service to check the ticket.

test/server_test.py
```diff
@@ -235,6 +254,23 @@ def started_imageio(tmpdir, drop_privileges="true"):
         proc.wait()
 
 
+@requires_root
+def test_load_ticket(tmpdir, tmp_dirs, tmp_image, ticket):
+    data = b"a" * 512 + b"b" * 512
+    tmp_image.write(data)
+    with started_imageio(tmpdir, drop_privileges="false", ticket=ticket):
```
nirs, line 261: why do we need to use drop_privileges="false" for this test? this feature should not depend on using initial ticket.

test/server_test.py
```diff
@@ -235,6 +254,23 @@ def started_imageio(tmpdir, drop_privileges="true"):
...
+        conf_dir = tmpdir.join("conf")
+        cfg = config.load(str(conf_dir.join("conf.d", "daemon.conf")))
+        with http.ControlClient(cfg) as c:
+            res = c.get('/tickets/file')
+            assert res.status == http_client.OK
+
+        with http.LocalClient(cfg) as c:
+            res = c.get('/images/file')
+            assert res.read() == data
+            assert res.status == http_client.OK
```
nirs, line 271: This makes sense, but why do need to use a local client instead of remote client? I guess this is the reason you tried to use a non-abstract unix socket for the configuration.

nirs (changes requested):
Too many changes, both related and unrelated.

This feature should not depend on anything else and change only the server module:

1. Add -t --ticket argument
2. If set, read ticket from specified path and add to server.auth
3. Add test starting a server with a ticket and verifying that the service provide the image specified in the ticket

aesteve-rh:
Thanks for the review! I'll address your comments and do some changes, but the branch is still WIP. I was still trying to find the best way to test it (that could also work in the CIs).

I'll clean it up today, and hopefully ready up the PR too.

test/server_test.py
aesteve-rh: I was not interested in running the non-existing ticket check if I am testing the initial ticket, and drop_privileges is always evaluated as true since it is a string.

test/server_test.py
aesteve-rh: It is awful checking 404 or 200, even when those values are generally well-known, when we have constants that tell us exactly what are we expecting.

Since I am adding more of these in my test, I thought would be nice to have them already.

test/server_test.py
aesteve-rh: Was failing in the CIs (and locally if run as root). This made it pass. But I am still doing changes with the test.

test/server_test.py
aesteve-rh: I get connection refused with remote client, currently investigating why.

test/server_test.py
aesteve-rh: Removed from the PR.

aesteve-rh:
I have left a few threads open for discussion, probably a few more tweaks needed. But is mature enough to ready it up.

test/server_test.py
nirs: You can use the constant in new tests. Improving unrelated tests to use constants is not required for this change, and mixing unrelated changes is bad.

test/server_test.py
aesteve-rh: Other tests mostly use the configuration contained in the Server instance, that has the obtained port updated. Since here we do not have access to the server instance, I had to set a specific port. But I used `testutil.random_tcp_port()` to obtain an unused one.

test/server_test.py
```diff
@@ -64,6 +67,20 @@ def tmp_dirs(tmpdir):
...
+def tmp_ticket(tmpdir):
+    img_size = 16 * KiB
+    img_path = testutil.create_tempfile(tmpdir, "disk.raw", size=img_size)
+    ticket_path = tmpdir.join("file.json")
+    ticket_path.write(
+        json.dumps(
+            testutil.create_ticket(
+                uuid="file",
+                size=img_size,
+                url=f'file://{img_path}')))
```
nirs, line 80: Trying to avoid temporary variables makes this code harder to read and write.

Why not:

    ticket = testutil.create_ticket(
        uuid="file", size=img_size, url=f'file://{img_path}')
    ticket.path.write(json.dumps(ticket)))

test/server_test.py
nirs: ok, this is a side effect of testing for missing ticket as a way to test that the server is responding.

We better change the test to test something else instead, since new we have new use case of starting a server with a ticket.

We can test instead `OPTIONS /images/*` or `GET /info` which does not depend on the control service or on having a ticket.

test/server_test.py
```diff
@@ -221,19 +241,34 @@ def started_imageio(tmpdir, drop_privileges="true"):
...
     finally:
         proc.terminate()
         proc.wait()
 
 
+def test_start_with_ticket(tmpdir, tmp_dirs, tmp_ticket):
+    data = b"a" * 8 * KiB + b"b" * 8 * KiB
+    with started_imageio(tmpdir, drop_privileges="false", ticket=tmp_ticket):
+        conf_dir = tmpdir.join("conf")
+        cfg = config.load(str(conf_dir.join("conf.d", "daemon.conf")))
```
nirs, line 261: Blank line after the steup paragraph and a comment before the check will make the test more clear:
   
    # Load configuration
    code...

    # Test that we can download.
    code...

    # Test that we can upload.
    code...

test/server_test.py
```diff
@@ -221,19 +241,34 @@ def started_imageio(tmpdir, drop_privileges="true"):
...
 
         yield proc
     finally:
         proc.terminate()
         proc.wait()
 
 
+def test_start_with_ticket(tmpdir, tmp_dirs, tmp_ticket):
+    data = b"a" * 8 * KiB + b"b" * 8 * KiB
+    with started_imageio(tmpdir, drop_privileges="false", ticket=tmp_ticket):
```
nirs, line 259: Once we fix the helper not to wait using the control socket, we can remove the drop_priveleges= arguemnt.

test/server_test.py
```diff
@@ -200,19 +217,22 @@ def prepare_config(tmpdir, drop_privileges="true"):
         run_dir=os.path.join(tmpdir, "run"),
         log_dir=os.path.join(tmpdir, "log"),
         drop_priv=drop_privileges,
+        remote_port=testutil.random_tcp_port(),
```
nirs, line 220: We should keep port=0 for all tests, using random_tcp_port() is needed only for starting a server with a ticket.

You can add a port=0 keyword arguemnt, and call it with a random_tcp_port() only in the specific test that needs a known port.

Another solution, test this feature in another test module starting the server in the same process, when we have access to the random port.

Testing here is problematic since this module uses the control socket for testing if the server is ready, and this does not work with initial ticket.

Testing the trivial code parsing the ticket argument is not very important, we can test that the Server class loads a ticket if ticket argument is specified.

nirs (changes requested):
Much better now, but I think this is the wrong place to test loading a ticket. Use another test module starting the server in the same process and keep this test as is.

aesteve-rh:
As discussed, added a new test module for containerized deployment.

Eventually (soonish), this will run tests with the actual imageio-ready container image.

In the pipelines, this should be tested separately, as it will employ a different container. Thus, it will probably require its own mark and tox environment. For now, though, we can keep it simple. Currently, it only tests the `--ticket` option with a Server instance.

ovirt_imageio/_internal/server.py
```diff
@@ -177,3 +184,17 @@ def _drop_privileges(self):
...
+        log.info("Initial ticket: %s", ticket)
+        self.auth.add(self._read_ticket(ticket))
+
+    def _read_ticket(self, ticket):
+        try:
+            with open(ticket, 'r') as f:
+                return json.loads(f.read())
+        except ValueError as e:
+            raise errors.InvalidTicket(
+                f"Ticket is not in a json format: {e}") from e
```
nirs, line 198: We don't know that, maybe:

    f"Cannot parse ticket {ticket}: {e}"

ovirt_imageio/_internal/server.py
```diff
@@ -177,3 +184,17 @@ def _drop_privileges(self):
...
+        except FileNotFoundError as e:
+            raise errors.InvalidTicket(f"Ticket not found: {e}") from e
```
nirs, line 200: The issue is missing file, so maybe:

    f"Cannot read ticket {ticket}: {e}"

(The ticket path may not be needed if it is included in e, I did not check)

ovirt_imageio/_internal/server.py
```diff
@@ -127,6 +131,9 @@ def __init__(self, config):
         if os.geteuid() == 0 and self.config.daemon.drop_privileges:
             self._drop_privileges()
 
+        if ticket:
+            self._add_ticket(ticket)
```
nirs, line 135: Loading the ticket after dropping privileges looks safe.

ovirt_imageio/_internal/server.py
```diff
@@ -44,7 +45,7 @@ def main():
         log.info("Starting (hostname=%s pid=%s, version=%s)",
                  socket.gethostname(), os.getpid(), version.string)
 
-        server = Server(cfg)
+        server = Server(cfg, args.ticket)
```
nirs, line 48: Using ticket=args.ticket is more clear and correct, since ticket is keyword argument. We always want to call with key=value style when using such arguments.

test/containerized_test.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+@pytest.fixture(
+    scope="module",
+    params=["daemon", "proxy"]
+)
+def srv_with_ticket(request, tmp_ticket):
+    path = "test/conf/{}.conf".format(request.param)
+    cfg = config.load(path)
+    s = server.Server(cfg, tmp_ticket)
```
nirs, line 49: ticket=tmp_ticket

test/containerized_test.py
```diff
@@ -0,0 +1,65 @@
...
+    ticket_path = tmpdir_factory.getbasetemp() / "file.json"
+    ticket = testutil.create_ticket(
+        uuid=TICKET_IMG_UUID, size=TICKET_IMG_SIZE, url=f'file://{tmp_image}')
+    ticket_path.write(json.dumps(ticket))
+    return str(ticket_path)
+
+
+@pytest.fixture(
+    scope="module",
+    params=["daemon", "proxy"]
```
nirs, line 44: The proxy config is not relevant for this test, so we can test only using the daemon conf.

test/containerized_test.py
```diff
@@ -0,0 +1,65 @@
...
+)
+def srv_with_ticket(request, tmp_ticket):
```
nirs, line 46: not sure we need _with_ticket - all the test here will use initial ticket, so we can use the common srv() name.

test/containerized_test.py
```diff
@@ -0,0 +1,65 @@
...
+
+from ovirt_imageio._internal import config
+from ovirt_imageio._internal import server
+from ovirt_imageio._internal.units import KiB
+
+from . import testutil
+from . import http
+
+
+TICKET_IMG_UUID = "file"
```
nirs, line 21: TICKET_ID = "ticket-id"?

But do we really need to use the same id? why not use random uuid for every ticket?

test/containerized_test.py
```diff
@@ -0,0 +1,65 @@
...
+TICKET_IMG_SIZE = 16 * KiB
```
nirs, line 22: TICKET_SIZE = 16 * KiB

test/containerized_test.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+@pytest.fixture(scope="module")
+def tmp_image(tmpdir_factory):
+    img_path = tmpdir_factory.getbasetemp() / "disk.raw"
```
nirs, line 27: Please use the style of the current code for creating temporary directories. Since we don't really need module scope, we can simply use the `tmpdir` fixture.

test/containerized_test.py
```diff
@@ -0,0 +1,65 @@
...
+
+from . import testutil
+from . import http
+
+
+TICKET_IMG_UUID = "file"
+TICKET_IMG_SIZE = 16 * KiB
+
+
+@pytest.fixture(scope="module")
```
nirs, line 25: Why do we need module scope? we have one test and starting it is pretty fast. Using the same ticket, image and server for all tests is not a good idea.

nirs (changes requested):

ovirt_imageio/_internal/server.py
aesteve-rh: I was trying to be consistent with the error given by the ticket Handler: https://github.com/oVirt/ovirt-imageio/blob/master/ovirt_imageio/_internal/handlers/tickets.py#L53

test/containerized_test.py
aesteve-rh: Since the fixture was in the module scope, I kept it unique. If I change it to a smaller scope, I will create random uuid per ticket.

test/containerized_test.py
aesteve-rh: I used it because it was used by the `srv` ficture that had the module scope.

test/containerized_test.py
aesteve-rh: I used the module scope so that it could be used from the server fixture, that had module scope as other tests do.

If it is ok to have new server per test, I'll remove all module fixtures and move to tmpdir instead of the factory.

test/containerized_test.py
```diff
@@ -0,0 +1,58 @@
...
+    s.start()
+    yield s
+    s.stop()
+
+
+@pytest.mark.parametrize("data", [
+    pytest.param(b"\0" * TICKET_SIZE, id="empty"),
+    pytest.param(
+        b"a" * (TICKET_SIZE // 2) + b"b" * (TICKET_SIZE // 2), id="full"),
+])
```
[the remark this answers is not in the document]
aesteve-rh: We do not need it. I added it because I was checking the validity of the fixtures, and wanted to do more tests, so I parametrized it. And after that I decided to leave it.

But you are right in that it does not add any verification value. Let's simplify it again.

test/containerized_test.py
```diff
@@ -0,0 +1,46 @@
...
+    with http.RemoteClient(srv.config) as c:
+        res = c.put(f'/images/{ticket_id}', data)
+        assert res.status == http_client.OK
+    # Test that we can download and matches the uploaded data
+    with http.RemoteClient(srv.config) as c:
+        res = c.get(f'/images/{ticket_id}')
+        assert res.read() == data
+        assert res.status == http_client.OK
+
+    srv.stop()
```
nirs, line 46: We need to stop in a try-finally block, otherwise when an assert fails this test will leave a running server.

nirs (changes requested):

test/containerized_test.py
aesteve-rh: True :facepalm:

nirs (approved):

PR #145, opened 2022-10-07
Containerized imageio deployment

aesteve-rh:
Start a container image for running imageio and container.

Using a container, the user will run an image transfer pod for every
transfer, exposing the pod for remote access, and connecting the
relevant storage the pod. When the transfer is done, the user will
unexpose the pod and terminate it.

In this mode the imageio server will serve one ticket, injected into the
pod when starting the pod, instead of adding a ticket using the control
service.

Signed-off-by: Nir Soffer [nsoffer@redhat.com](mailto:nsoffer@redhat.com)
Signed-off-by: Albert Esteve [aesteve@redhat.com](mailto:aesteve@redhat.com)

### DIFF
--- a/container/Containerfile
+++ b/container/Containerfile
@@ -15,4 +15,4 @@ RUN mkdir /etc/ovirt-imageio
 COPY --from=build /venv /venv
 COPY conf.d /etc/ovirt-imageio/conf.d
 EXPOSE 80/tcp
-CMD /venv/bin/ovirt-imageio
+CMD /venv/bin/ovirt-imageio -t ${TICKET_PATH}
--- a/container/Containerfile
+++ b/container/Containerfile
@@ -1,0 +1,18 @@
+# Build stage
+
+FROM alpine AS build
+RUN apk add --no-cache --update build-base linux-headers python3 python3-dev py3-pip
+WORKDIR /build
+COPY ovirt-imageio.tar.gz .
+RUN python3 -m venv /venv
+RUN /venv/bin/pip install ovirt-imageio.tar.gz
+
+# Final stage
+
+FROM alpine
+RUN apk add python3
+RUN mkdir /etc/ovirt-imageio
+COPY --from=build /venv /venv
+COPY conf.d /etc/ovirt-imageio/conf.d
+EXPOSE 80/tcp
+CMD /venv/bin/ovirt-imageio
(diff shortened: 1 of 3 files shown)


container/Containerfile
```diff
@@ -0,0 +1,20 @@
...
+COPY ovirt-imageio.tar.gz .
+RUN python3 -m venv /venv
+RUN /venv/bin/pip install ovirt-imageio.tar.gz
+
+# Final stage
+
+FROM alpine
+RUN apk add python3
+RUN addgroup --system ovirtimg
+RUN adduser --system --ingroup ovirtimg ovirtimg
```
nirs, line 15: We don't need to add the ovirtimg user and group since running as ovirtimg:ovirtimg
is not needed when running in a container and actually harmfull - when running as
ovirtimg we cannot access the mounted volumes since they are owned by root.

container/Containerfile
```diff
@@ -0,0 +1,20 @@
...
+RUN mkdir /run/ovirt-imageio /etc/ovirt-imageio
```
nirs, line 16: We don't need to create /run/ovirt-imageio since we can disable the control service
(if you rebase on #146).

container/conf.d/daemon.conf
```diff
@@ -0,0 +1,18 @@
+[daemon]
+systemd_enable = false
```
nirs, line 2: Add

    drop_privileges = false

So the server runs as root in the container - this fixes the permissions issue
that required running a privileged container.

container/conf.d/daemon.conf
```diff
@@ -0,0 +1,18 @@
...
+
+[tls]
+enable = false
+
+[remote]
+port = 80
+
```
nirs, line 9: Disable the control service:

    [control]
    enable = false

I tested these changes in
https://github.com/nirs/ovirt-imageio/commit/0a285564725f96f8f3baf56bb941cbaca6bd6fbe

nirs (changes requested):

nirs:
@aesteve-rh I pushed changes by mistake to your pr. If you don't see any
issue I think this is ready for merge.

nirs:
@aesteve-rh the example session from the commit message should
really be in docs/container.md, and later part of a tool making
this much easier to use.

aesteve-rh:
@nirs look good to me :) can you merge it (I cannot)?

Anyway, of all the changes, why disable the control service? is the only one I'm not sure why you do.

nirs (approved):

PR #146, opened 2022-10-08
Make control service optional

nirs:
When running in a container we will load the ticket using --ticket option so we don't need the control service.


aesteve-rh (approved):
lgtm!

PR #150, opened 2022-10-18
Fix push to quay.io

nirs:
Based on the examples[1] the quay.io username should be part of the `registry` value. Hopefully this will find the image built in the build step `localhost/image` and will push it to `quay.io/ovirt/image`.

[1] https://github.com/redhat-actions/push-to-registry#examples

### DIFF
--- a/.github/workflows/containers.yml
+++ b/.github/workflows/containers.yml
@@ -10,7 +10,7 @@ on:
   schedule:
     - cron:  '0 0 * * 1'
 env:
-  IMAGE_REGISTRY: quay.io
+  IMAGE_REGISTRY: quay.io/ovirt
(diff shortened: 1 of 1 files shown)


nirs:
Test run validating the fix:
https://github.com/nirs/ovirt-imageio/actions/runs/3275962455/jobs/5391448337

The push fails (expected) since I'm trying to push from my private fork, but the
previous error is fixed.

aesteve-rh (approved):
It's true that based on the action example the user should be on the registry, but I wonder why the Vdsm containers action is not failing in a similar way: https://github.com/oVirt/vdsm/blob/master/.github/workflows/containers.yml

Thanks for the fix, let's try again.

PR #152, opened 2022-10-20
container: serve image with nbd

aesteve-rh:
Include qemu-nbd in the container to make it
easier for users to use the containerized imageio
server.

The container now runs with a script as an
ENTRYPOINT. The script runs as a program that
can receive additional command line parameters.
The script is basically a launcher for the two
main services involved:
- qemu-nbd serving the image
- ovirt-imageio exposing the image through http

This commit removes support for file ticket, only
nbd tickets are implicitely used. This is better,
as nbd does much of the hard work for us, so will
perform better.

Users will not have to worry about
the ticket anyway, the ticket is also generated
automatically, so no need to mount the ticket
inside the container, or give an environment
variable. Users can set the ticket ID in the
command line, or use the file name as default.

Example session:
```
$ mkdir /var/tmp/images
$ qemu-img create -f raw /var/tmp/images/disk.raw 6g

$ podman run \
        --interactive \
        --tty \
        --rm \
        --publish 8080:80 \
        --volume /var/tmp/images:/var/tmp:Z \
        localhost/ovirt-imageio:latest /var/tmp/disk.raw
2022-10-20 14:46:08,953 INFO    (MainThread) [server] Starting (hostname=1594cc8a121b pid=8, version=2.4.7)
2022-10-20 14:46:08,970 INFO    (MainThread) [services] remote.service listening on ('::', 80)
2022-10-20 14:46:08,970 INFO    (MainThread) [server] Initial ticket: /ticket.json
2022-10-20 14:46:08,971 INFO    (MainThread) [server] Ready for requests
...
2022-10-26 12:07:00,155 INFO    (MainThread) [server] Received signal 2, shutting down

$ examples/imageio-client download -f raw http://localhost:8080/images/disk.raw download.raw
[ 100%  ] 6.00 GiB, 0.10 s, 57.23 GiB/s
```

### DIFF
--- a/container/Containerfile
+++ b/container/Containerfile
@@ -13,9 +13,10 @@ RUN /venv/bin/pip install ovirt-imageio.tar.gz
 # Final stage
 
 FROM alpine
-RUN apk add python3
+RUN apk add python3 qemu-img
 RUN mkdir /etc/ovirt-imageio
 COPY --from=build /venv /venv
 COPY conf.d /etc/ovirt-imageio/conf.d
+COPY app /app
 EXPOSE 80/tcp
-CMD /venv/bin/ovirt-imageio -t ${TICKET_PATH}
+ENTRYPOINT [ "/app/entrypoint.py" ]
--- a/container/app/entrypoint.py
+++ b/container/app/entrypoint.py
@@ -1,0 +1,67 @@
+#!/venv/bin/python3
+#
+# SPDX-FileCopyrightText: Red Hat, Inc.
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+import argparse
+import json
+import os
+import subprocess
+
+from ovirt_imageio._internal import qemu_img
+from ovirt_imageio._internal import qemu_nbd
+from ovirt_imageio._internal import nbd
+from ovirt_imageio._internal import util
+
+TICKET_PATH = "/app/ticket.json"
+OVIRT_IMAGEIO = "/venv/bin/ovirt-imageio"
+
+
+def main():
+    args = _parse_args()
+    image_info = qemu_img.info(args.image)
+    with util.tmp_dir("imageio-") as base:
+        socket_path = os.path.join(base, "sock")
+
+        ticket_info = _generate_ticket(
+            args.ticket_id, socket_path, image_info["virtual-size"])
+        with open(TICKET_PATH, "w") as f:
+            f.write(json.dumps(ticket_info))
+
+        sock = nbd.UnixAddress(socket_path)
+        with qemu_nbd.run(args.image, image_info["format"], sock):
+            subprocess.run(
+                [OVIRT_IMAGEIO, "--ticket", TICKET_PATH], check=True)
+
+
+def _generate_ticket(uuid, socket, size):
+    ticket_info = {}
+    ticket_info["uuid"] = uuid
+    ticket_info["url"] = f"nbd:unix:{socket}"
+    ticket_info["size"] = size
+    ticket_info["timeout"] = 3000
+    ticket_info["inactivity_timeout"] = 300
+    ticket_info["ops"] = ["read"]
+    return ticket_info
+
+
+def _parse_args():
+    parser = argparse.ArgumentParser(description="ovirt-imageio")
+    parser.add_argument("image", help="Path to image.")
+    parser.add_argument(
+        "--ticket-id",
+        help="Optional. Set the ID of the ticket. It is recommended "
+        "to use with random string for better security. "
+        "Defaults to the image filename.")
+
+    args = parser.parse_args()
+    # Ticket ID defaults to image filename if not set.
+    args.ticket_id = args.ticket_id or os.path.basename(args.image)
+    return args
+
+
+if __name__ == "__main__":
+    try:
+        main()
+    except KeyboardInterrupt:
+        print("Interrupted, shutting down")
--- a/ovirt_imageio/_internal/util.py
+++ b/ovirt_imageio/_internal/util.py
@@ -5,7 +5,11 @@ import collections
 import io
 import mmap
 import os
+import shutil
+import tempfile
 import threading
+
+from contextlib import contextmanager
 
 from .units import KiB
 
@@ -106,6 +110,15 @@ def ensure_text(s, encoding='utf-8', errors='strict'):
         return s
     else:
         raise TypeError("not expecting type '%s'" % type(s))
+
+
+@contextmanager
+def tmp_dir(prefix):
+    path = tempfile.mkdtemp(prefix=prefix)
+    try:
+        yield path
+    finally:
+        shutil.rmtree(path)
 
 
 class UnbufferedStream:
--- a/ovirt_imageio/client/_api.py
+++ b/ovirt_imageio/client/_api.py
@@ -8,10 +8,8 @@ api - imageio public client API.
 import json
 import logging
 import os
-import shutil
 import signal
 import tarfile
-import tempfile
 
 from contextlib import contextmanager
 from urllib.parse import urlparse
@@ -19,6 +17,7 @@ from urllib.parse import urlparse
 from .. _internal import blkhash
 from .. _internal import qemu_img
 from .. _internal import qemu_nbd
+from .. _internal import util
 from .. _internal.backends import http, nbd
 from .. _internal.handlers import checksum as _checksum
 from .. _internal.nbd import UnixAddress
@@ -489,7 +488,7 @@ def _open_nbd(filename, fmt, read_only=False, shared=1, bitmap=None,
     """
     Open nbd backend.
     """
-    with _tmp_dir("imageio-") as base:
+    with util.tmp_dir("imageio-") as base:
         sock = UnixAddress(os.path.join(base, "sock"))
 
         # If the applicatiion is handling signals, block SIGINT in qemu-nbd for
@@ -525,12 +524,3 @@ def _open_http(transfer_url, mode, cafile=None, secure=True, proxy_url=None):
                   transfer_url, e, proxy_url)
         url = urlparse(proxy_url)
         return http.open(url, mode, cafile=cafile, secure=secure)
-
-
-@contextmanager
-def _tmp_dir(prefix):
-    path = tempfile.mkdtemp(prefix=prefix)
-    try:
-        yield path
-    finally:
-        shutil.rmtree(path)


aesteve-rh:
First draft of the idea that we discussed for generating the ticket and serving with embedded nbd server in the container.

May need some tweaking. For example, maybe the `entrypoint.sh` can receive the ticket uuid from the user so that it does not need to copy it from the container output. But is a good starting point I feel.

container/bin/entrypoint.sh
```diff
@@ -0,0 +1,22 @@
...
+nbd-server.py $image_path &
+
+echo "Serving image with ID: $uuid"
+/venv/bin/ovirt-imageio -t $ticket_path &
+
+# Wait for any process to exit
+wait -n
+
+# Exit with status of process that exited first
+exit $?
```
nirs, line 22: Having one python script that does everything will be easier to understand and maintain. Now we have to use a shell script (with all the trouble of shell scripts), think about integrating with 2 additional tools.

Do you expect that the gen-ticket and nbd-server scripts have specific use cases
outside of the container?

container/bin/nbd-server.py
```diff
@@ -0,0 +1,41 @@
...
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+import argparse
+
+from ovirt_imageio.client import info
+from ovirt_imageio._internal import nbd
+from ovirt_imageio._internal import qemu_nbd
+
+
+DEFAULT_SOCKET_PATH = "/tmp/nbd.sock"
```
nirs, line 13: This is fine in a container, but unsafe outside of a container. The right way
is to create a temporary directory using mkdtemp() with a ranfom name and strict
permissions, so nobody can attack you using a symlink in /tmp/nbd.sock.

See how we do this in client._api. The best way to reuse that code is likely to move it to _internal and use it from both client and server scrtips.

nirs (changes requested):
The ticket-id should be set by the user, so they don't need to parse the output to send a request.

Or - can we use the disk image file name as the ticket uuid? While we call it uuid is is really a string. We cannot change the name now but we can introduce an alias like "name" instead and support both uuid and name (if uuid is not specied, use name).

nirs:
> $ curl http://localhost:8080/images/4e01dc7a-76e5-4533-bda5-24bacaddb82a -o /data/tmp/download.raw
> % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
>                                Dload  Upload   Total   Spent    Left  Speed
> 100 6144M  100 6144M    0     0  1342M      0  0:00:04  0:00:04 --:--:-- 1347M
> ```
> 
> [1] https://docs.docker.com/config/containers/multi-service_container/

Now that we support extents and we have #151 for using imageio-client, we really 
need to show download using the client.

nirs:
Tested on top of #151 
```
$ podman run         --interactive         --tty         --rm         --publish 8080:80         --volume /var/tmp/images:/var/tmp:Z         ovirt-imageio /var/tmp/disk.raw
Serving image with ID: 5cfbd26a-5f76-4dfa-9b0c-1c4c74743876
2022-10-20 22:45:43,465 INFO    (MainThread) [server] Starting (hostname=c1936b0d34a1 pid=8, version=2.4.7)
2022-10-20 22:45:43,475 INFO    (MainThread) [services] remote.service listening on ('::', 80)
2022-10-20 22:45:43,476 INFO    (MainThread) [server] Initial ticket: ticket.json
2022-10-20 22:45:43,476 INFO    (MainThread) [server] Ready for requests
2022-10-20 22:45:58,939 INFO    (Thread-1 (process_request_thread)) [http] OPEN connection=1 client=::ffff:10.0.2.100
2022-10-20 22:45:58,939 INFO    (Thread-1 (process_request_thread)) [nbd] Meta context qemu:allocation-depth is not available
2022-10-20 22:45:58,940 INFO    (Thread-1 (process_request_thread)) [images] [::ffff:10.0.2.100] OPTIONS transfer=(ticket/5cfbd26a-5f76-4dfa)
2022-10-20 22:45:58,940 INFO    (Thread-1 (process_request_thread)) [extents] [::ffff:10.0.2.100] EXTENTS transfer=(ticket/5cfbd26a-5f76-4dfa) context=zero
2022-10-20 22:45:58,976 INFO    (Thread-2 (process_request_thread)) [http] OPEN connection=2 client=::ffff:10.0.2.100
2022-10-20 22:45:58,979 INFO    (Thread-3 (process_request_thread)) [http] OPEN connection=3 client=::ffff:10.0.2.100
2022-10-20 22:45:58,980 INFO    (Thread-4 (process_request_thread)) [http] OPEN connection=4 client=::ffff:10.0.2.100
2022-10-20 22:45:58,980 INFO    (Thread-2 (process_request_thread)) [nbd] Meta context qemu:allocation-depth is not available
2022-10-20 22:45:58,980 INFO    (Thread-5 (process_request_thread)) [http] OPEN connection=5 client=::ffff:10.0.2.100
2022-10-20 22:45:58,983 INFO    (Thread-3 (process_request_thread)) [http] CLOSE connection=3 client=::ffff:10.0.2.100 [connection 1 ops, 0.004024 s]
2022-10-20 22:45:58,985 INFO    (Thread-5 (process_request_thread)) [http] CLOSE connection=5 client=::ffff:10.0.2.100 [connection 1 ops, 0.004834 s]
2022-10-20 22:45:58,986 INFO    (Thread-4 (process_request_thread)) [http] CLOSE connection=4 client=::ffff:10.0.2.100 [connection 1 ops, 0.005693 s]
2022-10-20 22:45:58,990 INFO    (Thread-2 (process_request_thread)) [http] CLOSE connection=2 client=::ffff:10.0.2.100 [connection 1 ops, 0.014222 s] [dispatch 1 ops, 0.008741 s] [read 1 ops, 0.007656 s, 16.00 KiB, 2.04 MiB/s] [read.read 1 ops, 0.007412 s, 16.00 KiB, 2.11 MiB/s] [read.write 1 ops, 0.000180 s, 16.00 KiB, 86.97 MiB/s]
2022-10-20 22:45:59,013 INFO    (Thread-1 (process_request_thread)) [http] CLOSE connection=1 client=::ffff:10.0.2.100 [connection 1 ops, 0.073618 s] [dispatch 2 ops, 0.001223 s] [extents 1 ops, 0.000275 s]
```

Downloading using curl:
```
$ curl http://localhost:8080/images/5cfbd26a-5f76-4dfa-9b0c-1c4c74743876 -o /data/tmp/download.raw
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 6144M  100 6144M    0     0  1197M      0  0:00:05  0:00:05 --:--:-- 1202M
```

Using imageio client:
```
$ examples/imageio-client download -f raw http://localhost:8080/images/5cfbd26a-5f76-4dfa-9b0c-1c4c74743876 /data/tmp/download.raw
[ 100% ] 6.00 GiB, 0.08 s, 79.13 GiB/s                                         
```

Trying a real image with 3g of data:

```
$ curl http://localhost:8080/images/d5c1c894-7278-4e8d-bbb8-854fd96ccd6e -o /data/tmp/download.raw
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 6144M  100 6144M    0     0   896M      0  0:00:06  0:00:06 --:--:--  842M

$ examples/imageio-client download -f raw http://localhost:8080/images/d5c1c894-7278-4e8d-bbb8-854fd96ccd6e /data/tmp/download.raw
[ 100% ] 6.00 GiB, 3.30 s, 1.82 GiB/s                                          
```

Awesome!

container/bin/entrypoint.sh
aesteve-rh: No, I changed my mind multiple times during the process, on how to structure the script. But I didn't think of specific uses outside the container, I was focused on the container.

Started with python entrypoint, actually. One script for all. But ended up with shell script coming from: https://docs.docker.com/config/containers/multi-service_container/
Also, I've seen this approach used in other repositories.

The main entrypoint script highlights pretty well what are the main services that the container runs in the background, and then waits for them. Makes the purpose clearer that if we do everything in one script. The main problem of it being a shell script in this context is that having more complex arguments is a bit of a pain, but since the idea was to keep the CLI simple, for now it works fine.

With respect to gen-ticket and `nbd-server` being separated, again, I did start them together. Then I felt like they were different things (mostly joint by the socket), and opted to split them, leaving the ticket generation as a helper script, and the `nbd-server` as an additional service run in the container. If we ever move that code to `ovirt-imageioctl`, I guess we would use only the `nbd-server` part anyway. But I'm open to change it.

container/bin/nbd-server.py
aesteve-rh: I didn't think to use it outside the container for now. But as we provide it, it may be tempting to use it. So I'll try to make it safer with your suggestion.

aesteve-rh:
I like the idea of using the image name as uuid. This can be a feature of ovirt-imageio, if a ticket does not have a uuid, default it to the file name (making the uuid optional in the `Ticket` class? -> https://github.com/oVirt/ovirt-imageio/blob/master/ovirt_imageio/_internal/auth.py#L24).

nirs:
> I like the idea of using the image name as uuid. This can be a feature of ovirt-imageio, if a ticket does not have a uuid, default it to the file name (making the uuid optional in the `Ticket` class? -> https://github.com/oVirt/ovirt-imageio/blob/master/ovirt_imageio/_internal/auth.py#L24).

Using file name is problematic. The ticket id is an authentication mechanism, and we don't
have a file name in all cases (e.g. unix socket). The server expects a string without 
"/" so we cannot use the full path of a unix socket, and using the filename does not ensure
unique ticket id (e.g. /tmp/xxxx/sock /tmp/yyyy/sock).

The issue is using "uuid" instead of "id", "token", or "name". We cannot change this since
existing code is assuming "uuid", but we can treat "uuid" as legacy name and support tickets
without uuid if they have the new name.

container/bin/entrypoint.py
```diff
@@ -0,0 +1,75 @@
...
+
+        gen_ticket(ticket_id, args.image, sock_path)
+        # Setup signals handler to silence backtrace on ctrl+C.
+        setup_signals()
+        sock = nbd.UnixAddress(sock_path)
+        img_format = info(args.image)["format"]
+        # Start services.
+        with qemu_nbd.run(args.image, img_format, sock) as server:
+            subprocess.Popen([OVIRT_IMAGEIO, "-t", TICKET_PATH])
+            server.wait()
```
nirs, line 71: This will wait until qemu-nbd exit, but it will never exit even if iamgeio exit.

We can use subprocess.run() to run imageio and wait until it exist, although it will never terminate:

    with qemu_nbd.run(...):
        subprocess.run([OVIRT_IMAGEIO, "--ticket", TICKET_PATH], check=True)

If imageio terminates, qemu-nbd will also terminate. Also lets always use long
options in code for better readability.

container/bin/entrypoint.py
```diff
@@ -0,0 +1,75 @@
...
+    with _tmp_dir("imageio-") as base:
+        # Use the random base directory if socket is not set.
+        sock_path = args.socket if args.socket else os.path.join(base, "sock")
+
+        gen_ticket(ticket_id, args.image, sock_path)
+        # Setup signals handler to silence backtrace on ctrl+C.
+        setup_signals()
+        sock = nbd.UnixAddress(sock_path)
+        img_format = info(args.image)["format"]
+        # Start services.
```
nirs, line 68: The comment is not really needed, the code is short and clear.

container/bin/entrypoint.py
```diff
@@ -0,0 +1,75 @@
...
+                 else os.path.basename(args.image))
+    with _tmp_dir("imageio-") as base:
+        # Use the random base directory if socket is not set.
+        sock_path = args.socket if args.socket else os.path.join(base, "sock")
+
+        gen_ticket(ticket_id, args.image, sock_path)
+        # Setup signals handler to silence backtrace on ctrl+C.
+        setup_signals()
+        sock = nbd.UnixAddress(sock_path)
+        img_format = info(args.image)["format"]
```
nirs, line 67: It is better to do this first - if we cannot access the image we should fail fast
before doing anything else. Current code is ok since podman cleans up anyway, but
it is more clear to do things in the right order regardless.

container/bin/entrypoint.py
```diff
@@ -0,0 +1,75 @@
...
+#
+# SPDX-FileCopyrightText: Red Hat, Inc.
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+import argparse
+import json
+import os
+import subprocess
+
+from ovirt_imageio.client import info
```
nirs, line 11: Do we really need client.info? This is a more fancy implementation that supports 
getting info from a file inside tar file, but here we want to do a basic 
qemu_img.info() call, so we should really import qemu_img instead.

Also importing functions from modules is bad, making it harder to understand the
code. Where is info coming from? Now we need to scroll up to check, and we may
shadow this name by a local variable.

container/bin/entrypoint.py
```diff
@@ -0,0 +1,75 @@
...
+from ovirt_imageio.client._api import _tmp_dir
```
nirs, line 12: Importing _tmp_dir from _api is ugly. We should not import private names!

To correct this, move _api._tmp_dir to another module as public name, like _util.tmp_dir.

nirs (changes requested):
Much nicer now! we have single file python file with does everything we need and it will be very easy to promote this to ovirt-imageioctl sub command if we want.

container/Containerfile
```diff
@@ -13,9 +13,10 @@ RUN /venv/bin/pip install ovirt-imageio.tar.gz
 # Final stage
 
 FROM alpine
-RUN apk add python3
+RUN apk add python3 qemu-img
 RUN mkdir /etc/ovirt-imageio
 COPY --from=build /venv /venv
 COPY conf.d /etc/ovirt-imageio/conf.d
+COPY bin /bin
```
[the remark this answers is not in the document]
aesteve-rh: I used bin to have the script available in the $PATH and invoke it without the full path.

In other project varies, many of them put it in the root folder, some in any other subfolder. But also the `bin` folder is used, e.g., https://docs.docker.com/develop/develop-images/dockerfile_best-practices/ (multi-stage builds section).

It is just a matter of being able to find the script later in the `ENTRYPOINT` line.

I don't have a preference nor think it makes a lot of difference. If you prefer `app`, is fine by me.

container/bin/entrypoint.py
```diff
@@ -0,0 +1,75 @@
...
+from ovirt_imageio.client._app import setup_signals
```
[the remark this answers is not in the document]
aesteve-rh: I just wanted to avoid the traceback in the output:
```
podman run \
        --interactive \
        --tty \
        --rm \
        --publish 8080:80 \
        --volume /var/tmp/images:/var/tmp:Z \
        localhost/ovirt-imageio:latest /var/tmp/disk.raw
2022-10-26 11:19:14,860 INFO    (MainThread) [server] Starting (hostname=2897d4c9e137 pid=12, version=2.4.7)
2022-10-26 11:19:14,873 INFO    (MainThread) [services] remote.service listening on ('::', 80)
2022-10-26 11:19:14,873 INFO    (MainThread) [server] Initial ticket: /ticket.json
2022-10-26 11:19:14,874 INFO    (MainThread) [server] Ready for requests
^C2022-10-26 11:19:16,785 INFO    (MainThread) [server] Received signal 2, shutting down
2022-10-26 11:19:16,816 INFO    (MainThread) [server] Received signal 15, shutting down
  File "/bin/entrypoint.py", line 76, in <module>
    main()
  File "/bin/entrypoint.py", line 67, in main
    server.wait()
  File "/venv/lib/python3.10/site-packages/ovirt_imageio/_internal/qemu_nbd.py", line 226, in wait
    return self.proc.wait(timeout)
  File "/usr/lib/python3.10/subprocess.py", line 1207, in wait
    return self._wait(timeout=timeout)
  File "/usr/lib/python3.10/subprocess.py", line 1941, in _wait
    (pid, sts) = self._try_wait(0)
  File "/usr/lib/python3.10/subprocess.py", line 1899, in _try_wait
    (pid, sts) = os.waitpid(self.pid, wait_flags)
KeyboardInterrupt
```

container/bin/entrypoint.py
```diff
@@ -0,0 +1,75 @@
...
+from ovirt_imageio._internal import nbd
+from ovirt_imageio._internal import qemu_nbd
+
+
+TICKET_PATH = "/ticket.json"
```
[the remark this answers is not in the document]
aesteve-rh: Only so that we have it in the `$PATH`, but placing it in the root folder could be a posibility, yes.

container/bin/entrypoint.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+def _generate_ticket(uuid, socket, size):
+    ticket_info = {}
+    ticket_info["uuid"] = uuid
+    ticket_info["url"] = f"nbd:unix:{socket}"
+    ticket_info["size"] = size
+    ticket_info["timeout"] = 3000
+    ticket_info["inactivity_timeout"] = 300
+    ticket_info["ops"] = ["read", "write"]
```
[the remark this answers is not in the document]
aesteve-rh: OK. I think we can add (later) a `--write` option or something like that to allow uploading. I'll open an Issue.

container/bin/entrypoint.py
```diff
@@ -0,0 +1,75 @@
...
+        f.write(json.dumps(ticket_info))
+
+
+def _parse_args():
+    parser = argparse.ArgumentParser(description="ovirt-imageio")
+    parser.add_argument("image", help="Path to image.")
+    parser.add_argument(
+        "-s", "--socket",
+        help="Optional. If set, ticket points to the given socket path. "
+        "Defaults to random temporary path.")
```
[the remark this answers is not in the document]
aesteve-rh: Not sure, I got the impression we wanted to support setting the socket. Maybe one could launch another process to the container attaching an interactive shell and be able to find the socket.

But I guess I got it wrong. If that's not an interesting usecase, then is for the better, supporting a defined socket path makes the script a bit more convoluted. I'll remove the option.

container/bin/entrypoint.py
```diff
@@ -0,0 +1,75 @@
...
+    # Ticket ID defaults to image filename if not set.
+    ticket_id = (args.ticket_id if args.ticket_id
+                 else os.path.basename(args.image))
+    with _tmp_dir("imageio-") as base:
+        # Use the random base directory if socket is not set.
+        sock_path = args.socket if args.socket else os.path.join(base, "sock")
+
+        gen_ticket(ticket_id, args.image, sock_path)
+        # Setup signals handler to silence backtrace on ctrl+C.
+        setup_signals()
```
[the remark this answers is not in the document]
aesteve-rh: I tried to catch `KeyboardInterrupt` and did not work. I think it catches the `KeyboardInterrupt` in the parent process, but the children (with `subprocess`) do not propagate, so cannot be caught.

I'll try again, though.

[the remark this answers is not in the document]
aesteve-rh: Now it worked, probably was the `wait` line, and since I removed the `wait` it does not fail anymore, so problem solved :)

container/Containerfile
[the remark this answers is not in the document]
nirs: Putting the entrypoint script in the path is not important, and it is better to run
it with absolute path anyway, so using bin/ is not required.

If we have multiple files, it makes sense to put all the files in the same directory
instead of adding random files at /. Since /app is documented, it will probably be
familiar to other developers, so it sounds like a good way.

I would copy the entrypoint script to /app, and also store the ticket int the same
place.

container/bin/entrypoint.py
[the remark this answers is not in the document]
nirs: Maybe --writable or --allow-write. Basically the opposite of --read-only.

ovirt_imageio/client/_utils.py
```diff
@@ -0,0 +1,20 @@
...
+"""
+
+import shutil
+import tempfile
+
+from contextlib import contextmanager
+
+
+@contextmanager
+def tmp_dir(prefix):
```
nirs, line 15: We already have _internal/util.py - why not add it there? This code is used both by the client (client._api) and server side (docker/entrypoint.py) so it should bn in the common package.

container/app/entrypoint.py
```diff
@@ -0,0 +1,70 @@
...
+        help="Optional. Set the ID of the ticket. It is recommended "
+        "to use with random string for better security. "
+        "Defaults to the image filename.")
+
+    return parser.parse_args()
+
+
+def main():
+    args = _parse_args()
+    image_format = qemu_img.info(args.image)["format"]
```
nirs, line 52: We need the image virtual-size when creating the ticket. You run qemu-img gain there, but this is not free and add unwanted delay. The right way is to run qemu-img once and pass the return value to the reset of the code.

We can do:

    image_info = image_format = qemu_img.info(args.image)
    ...
        create_ticket(ticket_id, image_info, socket_path)
        ...
        with qemu_nbd.run(args.image, image_info["format"], sock):

container/app/entrypoint.py
```diff
@@ -0,0 +1,70 @@
...
+    parser.add_argument(
+        "--ticket-id",
+        help="Optional. Set the ID of the ticket. It is recommended "
+        "to use with random string for better security. "
+        "Defaults to the image filename.")
+
+    return parser.parse_args()
+
+
+def main():
```
nirs, line 50: Lets keep the top-down approach used in _internal/server.py and client/_tool.py. The first function in the module is main.

container/app/entrypoint.py
```diff
@@ -0,0 +1,70 @@
...
+    ticket_info["size"] = size
+    ticket_info["timeout"] = 3000
+    ticket_info["inactivity_timeout"] = 300
+    ticket_info["ops"] = ["read"]
+    return ticket_info
+
+
+def create_ticket(ticket_id, image, socket):
+    image_size = qemu_img.info(image)["virtual-size"]
+    ticket_info = _generate_ticket(ticket_id, socket, image_size)
```
nirs, line 33: When passing image_info from main, we can use image_info["virtual-size"], so we don't need image_size.

Since the function will be just 3 lines and used once, maybe inline it into main.

container/app/entrypoint.py
```diff
@@ -0,0 +1,70 @@
...
+    with open(TICKET_PATH, "w+") as f:
```
nirs, line 34: Why do we need "w+"? We want to open the path, truncating the file if exists,
and write new content. The "w" mode does exactly this.

container/app/entrypoint.py
```diff
@@ -0,0 +1,70 @@
...
+    parser = argparse.ArgumentParser(description="ovirt-imageio")
+    parser.add_argument("image", help="Path to image.")
+    parser.add_argument(
+        "--ticket-id",
+        help="Optional. Set the ID of the ticket. It is recommended "
+        "to use with random string for better security. "
+        "Defaults to the image filename.")
+
+    return parser.parse_args()
+
```
nirs, line 48: We can implement the default value of args.ticket_id here. This is little cleaner
since we document the default here:

    args = parser.parse_args()
    args.ticket_id = args.ticket_id or os.path.basename(args.image)
    return args

Now we can pass args to the rest of the function since args.image and args.ticket_id are always complete. I think we use this same approach in ovirt-img
when we pass args initialized from both command line and config file. The rest of the application should not care how args was initialized.

nirs (changes requested):

container/app/entrypoint.py
aesteve-rh: At some point I'd say I tried, and it required also read access. I tested it now with `w` only and it works, so I am not sure why exactly did I add it. I'll leave it as `w`.

container/app/entrypoint.py
```diff
@@ -0,0 +1,67 @@
...
+        ticket_info = _generate_ticket(
+            args.ticket_id, socket_path, image_info["virtual-size"])
+        with open(TICKET_PATH, "w") as f:
+            f.write(json.dumps(ticket_info))
+
+        sock = nbd.UnixAddress(socket_path)
+        with qemu_nbd.run(args.image, image_info["format"], sock):
+            subprocess.run(
+                [OVIRT_IMAGEIO, "--ticket", TICKET_PATH], check=True)
```
nirs, line 34: Nice, we see everything in single function with mostly the same abstraction level.

nirs (approved):

PR #155, opened 2022-10-26
Support Ticket with name

aesteve-rh:
Current "uuid" key in the ticket dictionary
is misleading, as it does not necessarily
need to be an uuid, it is just a string to
identify the ticket.

We need to keep it for back-compatibility,
as other projects use the "uuid" key and
there is much code involved.

We can, instead, support "name" key in the
ticket dictionary, as an alternative to
"uuid". So we leave the "uuid" as a legacy
key, but still supported. In the ticket,
if not "uuid" nor "name" are set, we will
raise asking for "name", as is the new
key. If both are set, "uuid" prevails.

### DIFF
--- a/ovirt_imageio/_internal/auth.py
+++ b/ovirt_imageio/_internal/auth.py
@@ -21,7 +21,10 @@ class Ticket:
             raise errors.InvalidTicket(
                 "Invalid ticket: %r, expecting a dict" % ticket_dict)
 
-        self._uuid = _required(ticket_dict, "uuid", str)
+        self._uuid = _optional(ticket_dict, "uuid", str)
+        if self._uuid is None:
+            # Require 'name' and leave 'uuid' as a legacy key.
+            self._uuid = _required(ticket_dict, "name", str)
         self._size = _required(ticket_dict, "size", int)
         self._ops = _required(ticket_dict, "ops", list)
 
--- a/test/auth_test.py
+++ b/test/auth_test.py
@@ -384,6 +384,7 @@ def test_invalid_argument(arg, cfg):
 
 @pytest.mark.parametrize("kw", [
     {"uuid": 1},
+    {"name": 1},
     {"size": "not an int"},
     {"ops": "not a list"},
     {"timeout": "not an int"},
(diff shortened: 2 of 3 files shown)


ovirt_imageio/_internal/auth.py
```diff
@@ -21,7 +21,10 @@ def __init__(self, ticket_dict, cfg):
             raise errors.InvalidTicket(
                 "Invalid ticket: %r, expecting a dict" % ticket_dict)
 
-        self._uuid = _required(ticket_dict, "uuid", str)
+        self._uuid = _optional(ticket_dict, "uuid", str)
+        if self._uuid is None:
+            # Require 'name' and leave 'uuid' as a legacy key.
+            self._uuid = _required(ticket_dict, "name", str)
```
nirs, line 27: I would not keep the `_uuid` and `uuid` property - these are internal implementation
details and we can easy change them to `_name` and `name`. But maybe `_id` and `id` 
is a better term, since `name` does not have to be unique, but `id` is clearly is.

Regardless, changing clients to send the `name` instead of `uuid` is problematic
since vdsm is already assuming that a ticket has a `uuid`, and engine must support
old vdsm and imageio that require `uuid`. We can change engine to send both `name`
and `uuid` but I don't think such change is worth the effort.

We also use the uuid in the admin tool:

```
$ ./ovirt-imageioctl add-ticket -h
usage: ovirt-imageioctl add-ticket [-h] [-c CONF_DIR] filename

positional arguments:
  filename              Ticket filename.

options:
  -c CONF_DIR, --conf-dir CONF_DIR
                        Configuration directory (default /etc/ovirt-imageio).
```

We can change the tool to require a ticket-id, but having all the details
in the actual ticket json is nice, so I would not do this chagne.

If we change the `uuid` internally, and change the examples to match the new internal
name, we will have tickets cannot be used with ovirt (e.g. manually adding the ticket to vdsm). This is not great situation.

So I think this work simply does not worth the effort, and it is better to keep the
current state.

The new tool can use the term ticket_id instead of uuid, for example:

    entrypoint --ticket-id xxx-yyy /path/to/image/or/socket

nirs (commented):
This change is not complete, and it is not worth the effort. Lets keep this as is and make sure to use --ticekt-id in the entrypoint script, matching the actual api.

aesteve-rh:
Let's close it then. I was not adding the name property to keep the changes minimum and still support the new alias in the ticket. But I agree in that doing a more profound change is not worth the effort.

Issue #153, opened 2022-10-26
Add --insucure argument to ovirt-img


aesteve-rh:
Allow insecure connections using `ovirt-img` through a command line option.

nirs:
`client.download()` (and upload) already have `secure=True`, we just need to wire the
new option.

nirs:
@bennyz you should be able to use this soon.

PR #156, opened 2022-10-28
client: add --insecure option

aesteve-rh:
Allow insecure connections using ovirt-img through a command line option.

Fixes: #153

### DIFF
--- a/ovirt_imageio/client/_options.py
+++ b/ovirt_imageio/client/_options.py
@@ -46,12 +46,13 @@ log_level = Choices("log_level", ("debug", "info", "warning", "error"))
 
 
 class Option(
-        namedtuple("Option", "name,args,config,required,type,default,help")):
+        namedtuple(
+            "Option", "name,args,config,required,action,type,default,help")):
 
     __slots__ = ()
 
     def __new__(cls, name=None, args=(), config=False, required=False,
-                type=str, default=None, help=None):
+                action=None, type=str, default=None, help=None):
         """
         Arguments:
             name (str): Option name use in config file or parsed arguments.
@@ -61,6 +62,8 @@ class Option(
             required (bool): If True this option is required and parsing
                 arguments will fail if it is not specified in the command line
                 or config file.
+            action (Union[argparse.Action, str]): Specify how an argument
+                should be handled. Defaults to 'store' action.
             type (callable): Callable converting the command line argument or
                 config value to the wanted type. Must raise ValueError if the
                 value is invalid and cannot be converted.
@@ -69,7 +72,20 @@ class Option(
             help (str): Help message to show in the online help.
         """
         return tuple.__new__(
-            cls, (name, args, config, required, type, default, help))
+            cls, (name, args, config, required, action, type, default, help))
+
+    @property
+    def kwargs(self):
+        option_kwargs = {
+            "dest": self.name,
+            "default": None,
+            "help": self.help,
+        }
+        if self.action:
+            option_kwargs["action"] = self.action
+        else:
+            option_kwargs["type"] = self.type
+        return option_kwargs
 
 
 class Parser:
@@ -115,6 +131,14 @@ class Parser:
                   "from the specified config section"),
         ),
         Option(
+            name="secure",
+            args=["--insecure"],
+            action="store_false",
+            default=True,
+            help=("Do not verify server certificates and host name (not "
+                  "recommened).")
+        ),
+        Option(
             name="disk_timeout",
             config=True,
             type=int,
@@ -155,11 +179,7 @@ class Parser:
             if not option.args:
                 continue
 
-            cmd.add_argument(
-                *option.args,
-                dest=option.name,
-                type=option.type,
-                help=option.help)
+            cmd.add_argument(*option.args, **option.kwargs)
 
         if transfer_options:
             size = Size(minimum=1, default=MAX_WORKERS, maximum=8)
--- a/ovirt_imageio/client/_options.py
+++ b/ovirt_imageio/client/_options.py
@@ -20,6 +20,9 @@ import uuid
 
 
 ADD_DISK_TIMEOUT = 120
+# Possible boolean values in the configuration.
+BOOLEAN_VALUES = {'1': True, 'yes': True, 'true': True, 'on': True,
+                  '0': False, 'no': False, 'false': False, 'off': False}
 
 
 class Choices:
@@ -43,6 +46,12 @@ class Choices:
 
 
 log_level = Choices("log_level", ("debug", "info", "warning", "error"))
+
+
+def bool_string(value):
+    if value.lower() not in BOOLEAN_VALUES:
+        raise ValueError(f"Not a boolean: '{value}'")
+    return BOOLEAN_VALUES[value.lower()]
 
 
 class Option(
@@ -134,7 +143,9 @@ class Parser:
             name="secure",
             args=["--insecure"],
             action="store_false",
+            config=True,
             default=True,
+            type=bool_string,
             help=("Do not verify server certificates and host name (not "
                   "recommened).")
         ),
--- a/test/client_options_test.py
+++ b/test/client_options_test.py
@@ -154,6 +154,7 @@ def test_config_all(config):
     assert args.engine_url == "https://engine.com"
     assert args.username == "username"
     assert args.cafile == "/engine.pem"
+    assert args.secure is True
     assert args.disk_timeout == 200
     assert args.log_file == "/var/log/ovirt-img/engine.log"
     assert args.log_level == "info"
@@ -184,6 +185,7 @@ def test_config_all_override(config, tmpdir):
     assert args.engine_url == "https://engine2.com"
     assert args.username == "username2"
     assert args.cafile == "/engine2.pem"
+    assert args.secure is True
     assert args.disk_timeout == 200
     assert args.log_file == "test.log"
     assert args.log_level == "debug"
@@ -204,6 +206,7 @@ def test_config_required(config, monkeypatch):
     assert args.engine_url == "https://engine.com"
     assert args.username == "username"
     assert args.cafile is None
+    assert args.secure is True
     assert args.disk_timeout == 120
     assert args.log_file is None
     assert args.log_level == "warning"
@@ -226,12 +229,14 @@ def test_config_required_override(config, tmpdir):
         "-c", "required",
         "--password-file", str(password_file),
         "--cafile", "/engine2.pem",
+        "--insecure",
         "--log-file", "test.log",
         "--log-level", "debug",
     ])
     assert args.engine_url == "https://engine.com"
     assert args.username == "username"
     assert args.cafile == "/engine2.pem"
+    assert args.secure is False
     assert args.disk_timeout == 120
     assert args.log_file == "test.log"
     assert args.log_level == "debug"
--- a/test/client_options_test.py
+++ b/test/client_options_test.py
@@ -118,6 +118,7 @@ engine_url = https://engine.com
 username = username
 password = password
 cafile = /engine.pem
+secure = False
 disk_timeout = 200
 log_file = /var/log/ovirt-img/engine.log
 log_level = info
@@ -139,11 +140,17 @@ username = username
 #username is not set
 password = password
 
-[invalid]
+[invalid_log_level]
 engine_url = https://engine.com
 username = username
 password = password
-log_level = invalid log level
+log_level = invalid value
+
+[invalid_secure]
+engine_url = https://engine.com
+username = username
+password = password
+secure = invalid value
 """)
 
 
@@ -154,7 +161,7 @@ def test_config_all(config):
     assert args.engine_url == "https://engine.com"
     assert args.username == "username"
     assert args.cafile == "/engine.pem"
-    assert args.secure is True
+    assert args.secure is False
     assert args.disk_timeout == 200
     assert args.log_file == "/var/log/ovirt-img/engine.log"
     assert args.log_level == "info"
@@ -185,7 +192,7 @@ def test_config_all_override(config, tmpdir):
     assert args.engine_url == "https://engine2.com"
     assert args.username == "username2"
     assert args.cafile == "/engine2.pem"
-    assert args.secure is True
+    assert args.secure is False
     assert args.disk_timeout == 200
     assert args.log_file == "test.log"
     assert args.log_level == "debug"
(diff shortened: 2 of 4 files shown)


ovirt_imageio/client/_options.py
```diff
@@ -161,6 +161,14 @@ def add_sub_command(self, name, help, func, transfer_options=True):
                 type=option.type,
                 help=option.help)
 
+        cmd.add_argument(
+            "--insecure",
+            dest="secure",
+            action="store_false",
+            default=True,
+            help=("Do not verify server certificates and host name (not "
+                  "recommened)."))
```
nirs, line 170: Why not add this as Option like other options?

    Option(
        name="secure",
        args=["--insecure"],
        config=True,
        action="store_false",
        ...
    ),

This way we can control if the option can be set in the config file - you added
it to the config file, but I don't think the code will check it, since it does
not know about this option.

I guess we will have to add support for action= to the Options, but it should
be easy to add.

test/client_options_test.py
```diff
@@ -226,12 +229,14 @@ def test_config_required_override(config, tmpdir):
...
         "--log-file", "test.log",
         "--log-level", "debug",
     ])
     assert args.engine_url == "https://engine.com"
     assert args.username == "username"
     assert args.cafile == "/engine2.pem"
+    assert args.secure is False
     assert args.disk_timeout == 120
     assert args.log_file == "test.log"
     assert args.log_level == "debug"
```
nirs, line 242: We need to test the case when the config file contains secure=False, unless we don't
want to support this, but this seems useful for development environment, for closed networks where people do not care about the certificates.

nirs (changes requested):

ovirt_imageio/client/_options.py
aesteve-rh: I tried, added the support for action and everything, but `action` cannot coexist with `type` in the `add_argument` call:
```
            cmd.add_argument(
                *option.args,
                dest=option.name,
                action=option.action,
                type=option.type,
                help=option.help)
```
...
```
TypeError: __init__() got an unexpected keyword argument 'type'
``` 
I could make a dictionary, add the command arguments conditionally, and then pass the dictionary to `add_argument` as kwargs. But that felt like an over complication for the one single option (i.e., `secure`) that would use it.

nirs: Specifying action only when not None and type is bool using **kwargs dict should work.

ovirt_imageio/client/_options.py
```diff
@@ -243,7 +249,9 @@ def _merge_config(self, args):
 
             # Skip options set by the command line.
             if getattr(args, option.name, None) is not None:
-                continue
+                # These store actions create default values automatically.
+                if option.action not in ['store_false', 'store_true']:
+                    continue
```
nirs, line 254: Not very clear whats going on here. We used to skip non None option, meaning that the
user specified a value.

So now we skip these options only if option.action is not store_true or store_false.
Why?

ovirt_imageio/client/_options.py
```diff
@@ -45,6 +45,10 @@ def __repr__(self):
 log_level = Choices("log_level", ("debug", "info", "warning", "error"))
 
 
+def strtobool(str):
```
nirs, line 48: The name describes what this function does, but it is confusing when we use as a type.
Try to find another name that is more consistent with other types names in this module (e.g. log_level).

ovirt_imageio/client/_options.py
```diff
@@ -45,6 +45,10 @@ def __repr__(self):
...
+    return str in ['True', 'true', '1']
```
nirs, line 49: In conf file "Yes" and "yes" are also considered True. Check how configparser getbool() works. Can we use it to extract the value based on option.action?

ovirt_imageio/client/_options.py
```diff
@@ -46,12 +46,13 @@ def __repr__(self):
...
 class Option(
-        namedtuple("Option", "name,args,config,required,type,default,help")):
+        namedtuple(
+            "Option", "name,args,config,required,action,type,default,help")):
 
     __slots__ = ()
 
     def __new__(cls, name=None, args=(), config=False, required=False,
-                type=str, default=None, help=None):
+                action=None, type=str, default=None, help=None):
```
nirs, line 59: If the default is "store" (based on your comment bellow), can we use action="store"?

ovirt_imageio/client/_options.py
```diff
@@ -69,7 +72,18 @@ def __new__(cls, name=None, args=(), config=False, required=False,
...
+
+    def make_kwargs(self):
+        option_kwargs = {
+            "dest": self.name,
+            "help": self.help,
+        }
+        if self.action:
+            option_kwargs["action"] = self.action
+        else:
+            option_kwargs["type"] = self.type
```
nirs, line 89: Is this always correct or just special case of action="store_true" or action="store_false"?

ovirt_imageio/client/_options.py
```diff
@@ -69,7 +72,18 @@ def __new__(cls, name=None, args=(), config=False, required=False,
             help (str): Help message to show in the online help.
         """
         return tuple.__new__(
-            cls, (name, args, config, required, type, default, help))
+            cls, (name, args, config, required, action, type, default, help))
+
+    def make_kwargs(self):
```
nirs, line 81: Maybe implement this as read only property so we can do:

    cmd.add_argument(*options.args, **option.kwargs)

nirs (commented):
Nice!

ovirt_imageio/client/_options.py
aesteve-rh: The problem is that even if the user did not specify a value, if the variable has a default value, it will appear as a set (not None) option. The way we solved this, is by setting the default only **after** we read the configuration file options.

However, with 'store_false' and 'store_true' actions, default bool values are set automatically when we `parse_args` to `True` and `False` respectively, before processing the configuration file, thus preventing them to be processed from the config file. I have added the condition to avoid that, and making the file configuration to prevail over the command line option.

But for these two actions should be OK. In the configuration file, it only makes sense to set `secure=` option when set to `False` (as default is `True` already), and that cannot be made `secure=True` again from the command line anyway:

Config | CLI | Result
--- | --- | ---
Missing | Missing | True
Missing | False |  False
True(1) | * | True(1)
False | * | False

(1) This option does not make much sense, as default value is already True, no need to write it in the configuration.

ovirt_imageio/client/_options.py
aesteve-rh: It was an aclaratory comment, in `argparse`, `store` action is also the set from `None` (or 'store'):
```python
class _ActionsContainer(object):
...
        # register actions
        self.register('action', None, _StoreAction)
        self.register('action', 'store', _StoreAction)
```

Leaving it to `None` allows to easily distinguish when is set in the `Option` constructor, and when is left unset. So that we after do:
```python
        if self.action:
            option_kwargs["action"] = self.action
...
```

ovirt_imageio/client/_options.py
aesteve-rh: Just for the record, `getboolean` is implemented as a dictionary for possible states:
```python
    # Possible boolean values in the configuration.
    BOOLEAN_STATES = {'1': True, 'yes': True, 'true': True, 'on': True,
                      '0': False, 'no': False, 'false': False, 'off': False}
```
OK, I'll try to use it instead of setting the `type`.

ovirt_imageio/client/_options.py
nirs: Thanks, this is clear now. I think we don't want to use `action=store_xxx` since this
requires a default, and we this does not work with getting options both from the command
line and config.

We want to define the type in the option, and have `args.secure = None`. Then we can
set the default based on the config value (if set), the same way we treat other settings.

ovirt_imageio/client/_options.py
nirs: Using getboolean() is probably best, but I don't think it works with unset key.
Using get() we can detect if we have a value, and then convert the value to 
bool using the same rules (maybe using the dict you show here).

ovirt_imageio/client/_options.py
nirs: Ok, probably not relevant if we don't use `action=store_xxx`.

ovirt_imageio/client/_options.py
nirs: Try to avoid support for action= will probably be best.

aesteve-rh:
Checking how it looks with a custom `Action`.

ovirt_imageio/client/_options.py
```diff
@@ -45,13 +45,34 @@ def __repr__(self):
 log_level = Choices("log_level", ("debug", "info", "warning", "error"))
 
 
+class StoreFalseOrNoneAction(argparse.Action):
```
nirs, line 48: StoreFalseOrNone is good enough. The Action thing come from the super class, and when
use we always use:

    action=StoreFalseOrNone(...)

So it is also clear.

ovirt_imageio/client/_options.py
```diff
@@ -45,13 +45,34 @@ def __repr__(self):
...
+    some sort of flag.
+    """
+
+    def __init__(self, option_strings, dest, required=False, help=None):
+        super().__init__(
+            option_strings=option_strings,
+            dest=dest,
+            nargs=0,
+            required=required,
+            help=help)
```
nirs, line 62: If we just pass arguments to the super constructor, we usually don't need to implement
`__init__`.

Looks like this version passes option_strings, dest, required, and help, and use nargs=0 as default. Don't we get the same result if we don't implement `__init__`?

ovirt_imageio/client/_options.py
```diff
@@ -45,13 +45,34 @@ def __repr__(self):
...
+
+    def __call__(self, parser, namespace, values, option_string=None):
+        setattr(namespace, self.dest, False)
```
nirs, line 65: Is this called only when user specifies --insecure?

ovirt_imageio/client/_options.py
```diff
@@ -267,11 +268,14 @@ def _merge_config(self, args):
 
             # Read and validate value from config file.
             if config.has_option(args.config, option.name):
-                raw_value = config.get(args.config, option.name)
-                try:
-                    value = option.type(raw_value)
-                except ValueError as e:
-                    self._parser.error(e)
+                if option.action is StoreFalseOrNoneAction:
+                    value = config.getboolean(args.config, option.name)
```
nirs, line 272: This is a bit ugly, depending on the type of the action. Can we use the type of the
default instead?

    if type(option.default) is bool:

But the best would be if the common code works for everything:

    value = option.type(raw_value)

And we implement converting from string to bool in our code instead of using
config.getboolean().

ovirt_imageio/client/_options.py
aesteve-rh: By default, `nargs` takes a single value, but we want it to explicitly be `0`.

ovirt_imageio/client/_options.py
aesteve-rh: Yeah, I was not convinced by this either. I tried with nested `try...except`, but the checks were misbehaving.

I guess that I will go back to the type doing the conversion.

ovirt_imageio/client/_options.py
aesteve-rh: Yes, I think so.

ovirt_imageio/client/_options.py
nirs: Add

   "default": None

So ensure that `store_false` will not set the default value.

ovirt_imageio/client/_options.py
nirs: We can simplify this a lot like this:

```
$ cat opt.py 
import argparse

p = argparse.ArgumentParser()
p.add_argument("--insecure", dest="secure", action='store_false', default=None)
args = p.parse_args()
print(args)

$ python opt.py --insecure
Namespace(secure=False)

$ python opt.py
Namespace(secure=None)
```

So the only thing to do to make this work is to pass

    default=None

when using store_xxx actions, because these action classes uses:

```
class _StoreFalseAction(_StoreConstAction):

    def __init__(self,
                 option_strings,
                 dest,
                 default=True,
                 required=False,
                 help=None):
        super(_StoreFalseAction, self).__init__(
            option_strings=option_strings,
            dest=dest,
            const=False,
            default=default,
            required=required,
            help=help)
```

Since we never want to set the default using argparse, we we need to add

    "default": None

To our new shiny kwargs() property.

nirs (changes requested):

nirs:
Suggested changes tested here:
https://github.com/nirs/ovirt-imageio/commit/064ac922d974f94d3575528b680dac15257df5c0

ovirt_imageio/client/_options.py
```diff
@@ -20,6 +20,9 @@
 
 
 ADD_DISK_TIMEOUT = 120
+# Possible boolean values in the configuration.
+BOOLEAN_STATES = {'1': True, 'yes': True, 'true': True, 'on': True,
```
nirs, line 24: BOOLEAN_VALUES?

test/client_options_test.py
```diff
@@ -139,11 +140,17 @@ def config(tmpdir, monkeypatch):
 #username is not set
 password = password
 
-[invalid]
+[invalid1]
```
nirs, line 143: invalid_log_level?

test/client_options_test.py
```diff
@@ -139,11 +140,17 @@ def config(tmpdir, monkeypatch):
...
 engine_url = https://engine.com
 username = username
 password = password
 log_level = invalid log level
+
+[invalid2]
```
nirs, line 149: invalid_secure?

test/client_options_test.py
```diff
@@ -139,11 +140,17 @@ def config(tmpdir, monkeypatch):
...
+engine_url = https://engine.com
+username = username
+password = password
+secure = invalid value
```
nirs, line 153: for consistency with the other invalid config:

    secure = invalid secure

Or just use 'invalid value' in both to simplify. This will also make the test simpler, we will not need the value argument since the value is always the same.

nirs (changes requested):

ovirt_imageio/client/_options.py
aesteve-rh: I realized the `default=True` in the store Action, but didn't think about leveraging the `kwargs` option to default it to `None` and then set it later. That's a nice trick!

nirs (approved):

PR #166, opened 2022-11-18
Switch from Python 3.8 to 3.9 on EL8

mwperina:
Switch from Python 3.8 to 3.9 on EL8 to be able to support
ansible-core-2.13

Signed-off-by: Martin Perina <mperina@redhat.com>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
           - distro: centos-8
             env: py36
           - distro: centos-8
-            env: py38
+            env: py39
           - distro: centos-9
             env: py39
           - distro: fedora-36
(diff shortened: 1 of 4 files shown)


aesteve-rh (approved):

.github/workflows/ci.yml
```diff
@@ -17,7 +17,7 @@ jobs:
           - distro: centos-8
             env: py36
           - distro: centos-8
-            env: py38
+            env: py39
```
aesteve-rh, line 20: As we are removing Python 3.8 from the CIs (and technically dropping support), we could remove `*-py38` from the available `tox` environments. Let's keep it for now and decide if we want to remove it in a follow-up patch.

PR #167, opened 2022-11-20
qemu_nbd: add support for block devices

bennyz:
In kubernetes a PVC may have a raw block volume mode, in which case we'd need to write to a block device and not to a file.

This patch checks whether the provided path is a block device and uses the appropriate driver, as well as set the cache to writeback since aio=native doesn't work with an RBD mapped blocked device.

### DIFF
--- a/ovirt_imageio/_internal/qemu_nbd.py
+++ b/ovirt_imageio/_internal/qemu_nbd.py
@@ -7,6 +7,7 @@ import logging
 import os
 import signal
 import subprocess
+import stat
 import urllib.parse
 
 from contextlib import contextmanager
@@ -89,7 +90,11 @@ class Server:
     def start(self):
         # Implement "auto" cache mode, planned for qemu 7.
         if self.cache is None:
-            self.cache = "none" if self._can_use_direct_io() else "writeback"
+            if self._can_use_direct_io() or self._is_block_dev():
+                self.cache = "writeback"
+            else:
+                self.cache = "none"
+
             log.debug("Using cache=%r", self.cache)
 
         if self.aio is None:
@@ -147,7 +152,10 @@ class Server:
         # Build a 'json:{...}' filename allowing control all aspects of the
         # image.
 
-        file = {"driver": "file", "filename": self.image}
+        if self._is_block_dev():
+            file = {"driver": "host_device", "filename": self.image}
+        else:
+            file = {"driver": "file", "filename": self.image}
 
         if self.offset is not None or self.size is not None:
             # Exposing a range in a raw file.
@@ -245,6 +253,11 @@ class Server:
         if self.block_signals:
             signal.pthread_sigmask(signal.SIG_BLOCK, self.block_signals)
 
+    def _is_block_dev(self):
+        """
+        Checks if a path is of a block device
+        """
+        return stat.S_ISBLK(os.lstat(self.image).st_mode)
 
 @contextmanager
 def run(image, fmt, sock, export_name="", read_only=False, shared=8,


nirs:
> In kubernetes a PVC may have a raw block volume mode, in which case we'd need to write to a block device and not to a file.
> 
> This patch checks whether the provided path is a block device and uses the appropriate driver, as well as set the cache to writeback since aio=native doesn't work with an RBD mapped blocked device.

The patch disables cache=none for any block device which does not make sense. If we have
an issue with RBD mapped block device (what is this?) we need a more specific solution.

Is this this kind of device?
https://docs.ceph.com/en/latest/rbd/rbd-kubernetes/

ovirt_imageio/_internal/qemu_nbd.py
```diff
@@ -147,7 +152,10 @@ def start(self):
         # Build a 'json:{...}' filename allowing control all aspects of the
         # image.
 
-        file = {"driver": "file", "filename": self.image}
+        if self._is_block_dev():
+            file = {"driver": "host_device", "filename": self.image}
+        else:
+            file = {"driver": "file", "filename": self.image}
```
nirs, line 158: This can be simpler:

    driver = "host_device" if self._is_block_dev() else "file"
    file = {"driver": driver, "filename": self.image}

But calling os.stat() multiple times does not make sense. We should really call it once when starting and set a flag.

ovirt_imageio/_internal/qemu_nbd.py
```diff
@@ -245,6 +253,11 @@ def _preexec_fn(self):
         if self.block_signals:
             signal.pthread_sigmask(signal.SIG_BLOCK, self.block_signals)
 
+    def _is_block_dev(self):
+        """
+        Checks if a path is of a block device
+        """
+        return stat.S_ISBLK(os.lstat(self.image).st_mode)
```
nirs, line 260: os.lstat() does not follow symbolic links, which is the worng thing to do since /dev/xxx may be symlink to the real device (e.g. /dev/dm-42). os.stat() is the right function.

bennyz:
> > In kubernetes a PVC may have a raw block volume mode, in which case we'd need to write to a block device and not to a file.
> > This patch checks whether the provided path is a block device and uses the appropriate driver, as well as set the cache to writeback since aio=native doesn't work with an RBD mapped blocked device.
> 
> The patch disables cache=none for any block device which does not make sense. If we have an issue with RBD mapped block device (what is this?) we need a more specific solution.
> 
> Is this this kind of device? https://docs.ceph.com/en/latest/rbd/rbd-kubernetes/

I mapped it locally with:
```shell
$ rbd map volumes/hello
```

I guess on kubernetes it uses rbd-nbd

ovirt_imageio/_internal/qemu_nbd.py
```diff
@@ -89,7 +90,11 @@ def url(self):
     def start(self):
         # Implement "auto" cache mode, planned for qemu 7.
         if self.cache is None:
-            self.cache = "none" if self._can_use_direct_io() else "writeback"
+            if self._can_use_direct_io() or self._is_block_dev():
+                self.cache = "writeback"
+            else:
+                self.cache = "none"
```
nirs, line 96: This is wrong, will change self.cache=None to self.cache="writeback" for file that can
use direct I/O.

But even if we fix it to:

    if self._can_use_direct_io() or not self._is_block_dev():
        self.cache = "none"
    else:
        self.cache = "writeback"

This is still not correct since block devices typically support direct I/O
and if direct I/O is not supported opening the image using direct I/O should
fail. Vdsm was using this kind of check for 10 years and we never found a case
when this did not work. But maybe the check need to be improved, and we need
to try to read from the device instead of just opening it?

Can you share information from your setup?

On the host:

    kubectl get pvc/xxx ... -o yaml
    kubectl get pv/xxx ... -o yaml

In the container:

    udevadm info /dev/xxx
    dd if=/dev/urandom bs=1m count=1 of=/var/tmp/data
    qemu-img convert -f raw -O raw -t none /var/tmp/data /dev/xxx

nirs (changes requested):

nirs:
> I mapped it locally with:
> 
> ```shell
> $ rbd map volumes/hello
> ```
> 
> I guess on kubernetes it uses rbd-nbd

Why do this manually?

What happens if you create a pvc as explained in
https://docs.ceph.com/en/latest/rbd/rbd-kubernetes/#create-a-persistentvolumeclaim

nirs:
I tested direct I/O read/write to ceph block volume like this:

1. Start minikube cluster using rook ceph block pool:
```
$ cat rook-block.yaml
---
profiles:
  - name: "rook-block"
    network: default
    memory: "6g"
    extra_disks: 1
    disk_size: "50g"
    - scripts:
        - file: rook-operator/start
        - file: rook-cluster/start
        - file: rook-toolbox/start

$ drenv start rook-block.yaml
```
See https://github.com/RamenDR/ramen/tree/main/test for the details.

2. Add rbd-block-pvc:
```
$ cat rbd-block-pvc.yaml 
---
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: rbd-block-pvc
spec:
  accessModes:
    - ReadWriteOnce
  volumeMode: Block
  resources:
    requests:
      storage: 1Gi
  storageClassName: rook-ceph-block

$ kubectl apply -f rook-block-pvc.ymal

$ kubectl get pvc/rbd-block-pvc
NAME            STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS      AGE
rbd-block-pvc   Bound    pvc-7478ef0e-de21-4fe0-9aeb-d5fc992f9ddd   1Gi        RWO            rook-ceph-block   56m
```

3. Create pod using the pvc
```
$ cat dio-pod.yaml 
---
apiVersion: v1
kind: Pod
metadata:
  name: dio-test
spec:
  containers:
    - name: dio-test
      image: busybox
      command: ["sleep", "600"]
      volumeDevices:
        - name: data
          devicePath: /dev/xvda
  volumes:
    - name: data
      persistentVolumeClaim:
        claimName: rbd-block-pvc
```

3. Test that the pod has a block device
```
$ kubectl exec pod/dio-test -it -- stat /dev/xvda
  File: /dev/xvda
  Size: 0         	Blocks: 0          IO Block: 4096   block special file
Device: 200006h/2097158d	Inode: 12          Links: 1     Device type: fb,0
Access: (0660/brw-rw----)  Uid: (    0/    root)   Gid: (    6/    disk)
Access: 2022-11-20 15:10:23.307487166 +0000
Modify: 2022-11-20 15:10:23.307487166 +0000
Change: 2022-11-20 15:10:23.307487166 +0000
```

4. Test writing using direct I/O
```
$ kubectl exec pod/dio-test -it -- dd if=/dev/zero bs=1M count=1 of=/dev/xvda oflag=direct conv=fsync,notrunc
1+0 records in
1+0 records out
1048576 bytes (1.0MB) copied, 0.010783 seconds, 92.7MB/s
```

5. Test reading using direct I/O
```
$ kubectl exec pod/dio-test -it -- dd of=/dev/null bs=1M count=1 if=/dev/xvda iflag=direct
1+0 records in
1+0 records out
1048576 bytes (1.0MB) copied, 0.002711 seconds, 368.9MB/s
```

So using direct I/O with ceph block device works, and cache=none should work.

bennyz:
> I tested direct I/O read/write to ceph block volume like this:
> 
> 1. Start minikube cluster using rook ceph block pool:
> 
> ```
> $ cat rook-block.yaml
> ---
> profiles:
>   - name: "rook-block"
>     network: default
>     memory: "6g"
>     extra_disks: 1
>     disk_size: "50g"
>     - scripts:
>         - file: rook-operator/start
>         - file: rook-cluster/start
>         - file: rook-toolbox/start
> 
> $ drenv start rook-block.yaml
> ```
> 
> See [https://github.com/nirs/ramen/tree/rook-split/test](https://github.com/nirs/ramen/tree/rook-split/test?rgh-link-date=2022-11-20T15%3A19%3A25Z) (using rook-split branch)
> 
> 2. Add rbd-block-pvc:
> 
> ```
> $ cat rbd-block-pvc.yaml 
> ---
> apiVersion: v1
> kind: PersistentVolumeClaim
> metadata:
>   name: rbd-block-pvc
> spec:
>   accessModes:
>     - ReadWriteOnce
>   volumeMode: Block
>   resources:
>     requests:
>       storage: 1Gi
>   storageClassName: rook-ceph-block
> 
> $ kubectl apply -f rook-block-pvc.ymal
> 
> $ kubectl get pvc/rbd-block-pvc
> NAME            STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS      AGE
> rbd-block-pvc   Bound    pvc-7478ef0e-de21-4fe0-9aeb-d5fc992f9ddd   1Gi        RWO            rook-ceph-block   56m
> ```
> 
> 3. Create pod using the pvc
> 
> ```
> $ cat dio-pod.yaml 
> ---
> apiVersion: v1
> kind: Pod
> metadata:
>   name: dio-test
> spec:
>   containers:
>     - name: dio-test
>       image: busybox
>       command: ["sleep", "600"]
>       volumeDevices:
>         - name: data
>           devicePath: /dev/xvda
>   volumes:
>     - name: data
>       persistentVolumeClaim:
>         claimName: rbd-block-pvc
> ```
> 
> 3. Test that the pod has a block device
> 
> ```
> $ kubectl exec pod/dio-test -it -- stat /dev/xvda
>   File: /dev/xvda
>   Size: 0         	Blocks: 0          IO Block: 4096   block special file
> Device: 200006h/2097158d	Inode: 12          Links: 1     Device type: fb,0
> Access: (0660/brw-rw----)  Uid: (    0/    root)   Gid: (    6/    disk)
> Access: 2022-11-20 15:10:23.307487166 +0000
> Modify: 2022-11-20 15:10:23.307487166 +0000
> Change: 2022-11-20 15:10:23.307487166 +0000
> ```
> 
> 4. Test writing using direct I/O
> 
> ```
> $ kubectl exec pod/dio-test -it -- dd if=/dev/zero bs=1M count=1 of=/dev/xvda oflag=direct conv=fsync,notrunc
> 1+0 records in
> 1+0 records out
> 1048576 bytes (1.0MB) copied, 0.010783 seconds, 92.7MB/s
> ```
> 
> 5. Test reading using direct I/O
> 
> ```
> $ kubectl exec pod/dio-test -it -- dd of=/dev/null bs=1M count=1 if=/dev/xvda iflag=direct
> 1+0 records in
> 1+0 records out
> 1048576 bytes (1.0MB) copied, 0.002711 seconds, 368.9MB/s
> ```
> 
> So using direct I/O with ceph block device works, and cache=none should work.

I think the problem is with qemu rather than with the device itself, this is what I get:
```shell
2022-11-20 18:13:36,553 WARNING (MainThread) [qemu_nbd] qemu-nbd failed rc=1 err=b'qemu-nbd: Failed to blk_new_open \'json:{"driver": "raw", "file": {"driver": "host_device", "filename": "/dev/rbd0"}}\': aio=native was specified, but is not supported in this build.\n'
```
I have qemu-img-7.0.0-11.fc37.x86_64

nirs:
I added #168 for this, this also affects the ovirt-imageio container.

nirs:
@bennyz do we need still need this?

nirs:
No reply from @bennyz, time to close.

PR #169, opened 2022-11-20
Support block devices in qemu_nbd

nirs:
Add support for block devices in qemu_nbd module. This enables serving block devices
with the ovirt-imageio container and writing to block devices in the ovirt-img tool.

Based on #167, adding basic tests.

### DIFF
--- a/ovirt_imageio/_internal/qemu_nbd.py
+++ b/ovirt_imageio/_internal/qemu_nbd.py
@@ -6,6 +6,7 @@ import json
 import logging
 import os
 import signal
+import stat
 import subprocess
 import urllib.parse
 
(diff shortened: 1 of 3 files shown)


aesteve-rh (approved):
Nice!

nirs:
> This stales #167.
> 
> Is there anything left to do with it or we can close it after we merge this?

#167 include this change (without the tests) but disables direct i/o for block devices (testing show that this is not needed).

@bennyz need to test this change, if this works we can close #167.

bennyz:
It works remotely for me, still fails locally aio=native but I can live with it

bennyz (approved):

bennyz:
> > It works remotely for me, still fails locally aio=native but I can live with it
> 
> @bennyz What do you mean by remotely and locally? I want to know if this works in the actual openshift environment int the volume populator pod.

remote = openshift
local = my laptop

PR #173, opened 2022-12-19
containers: create venv for tests

aesteve-rh:
In test containers, create a
new venv [1] at /venv instead of installing
the required packages directly in the
root environment. Installing packages in
the root environment is strongly discouraged,
as it can run into conflicts with the system
package manager.

[1] - https://pip.pypa.io/warnings/venv


aesteve-rh:
Failed job with conflict with the system package manager: https://github.com/oVirt/ovirt-imageio/actions/runs/3727372549/jobs/6321458142

nirs (approved):

PR #177, opened 2023-02-15
ci test: activate venv before tests

aesteve-rh:
Activate containers venv before running
tests in the pipeline.

### DIFF
--- a/ci/test.sh
+++ b/ci/test.sh
@@ -32,5 +32,8 @@ fi
 
 env="$1"
 
+# Activate the tests venv
+source /venv/bin/activate
+
 make storage
 tox -e reuse,flake8,test-$env,bench-$env
--- a/tox.ini
+++ b/tox.ini
@@ -21,9 +21,8 @@ deps =
     test: pytest-cov
     test: pytest-timeout
 commands_pre =
-    # Build ioutil for both current python (e.g. 3.8) and platform python (e.g. 3.6)
+    # Build ioutil for current python (e.g. 3.8)
     python setup.py build_ext --build-lib .
-    /usr/bin/python3 setup.py build_ext --build-lib .
     python -c 'from test import testutil; print("ipv6 supported: %s" % testutil.ipv6_enabled())'
 commands =
     test: pytest -m 'not benchmark' --cov=ovirt_imageio --durations=10 {posargs}


ci/test.sh
```diff
@@ -32,5 +32,8 @@ fi
 
 env="$1"
 
+# Activate the tests venv
+source /venv/bin/activate
```
nirs, line 36: This makes sense if we installed the requirements in the virtual env.

I wonder how the tests passed before this change?

tox.ini
```diff
@@ -21,9 +21,8 @@ deps =
     test: pytest-cov
```
nirs, line 21: Replace `ioctl` with `ioutil` in the commit message.

tox.ini
```diff
@@ -21,9 +21,8 @@ deps =
...
     test: pytest-timeout
 commands_pre =
-    # Build ioutil for both current python (e.g. 3.8) and platform python (e.g. 3.6)
+    # Build ioutil for current python (e.g. 3.8)
```
nirs, line 24: There was a reason for building ioutil for both current and platform python. Please find what was the reason in git history before removing this, and explain why it is ok to do this now.

tox.ini
```diff
@@ -21,9 +21,8 @@ deps =
...
     python setup.py build_ext --build-lib .
-    /usr/bin/python3 setup.py build_ext --build-lib .
```
nirs, line 26: This fail with an error suggesting to include this version in allow_externals, so this will work also when runing inside the virtual environment. Maybe this is a better solution?

Or maybe this is not needed now since we change the way we run the tests?

Can you run the tests locally with this change?

ci/test.sh
aesteve-rh: I guess we haven't tested the pipelines since the update.
Crearly, I forgot to update this project, as I was doing the same for `vdsm` in parallel.

tox.ini
aesteve-rh: I'll do.

For the record, this is the commit: https://github.com/oVirt/ovirt-imageio/commit/b76751afaf3e260b638a501458be3d61912dfa6e

tox.ini
aesteve-rh: Yes, I can run them locally.

The problem seemed to be that `python` and `/usr/bin/python` pointed to different Python versions (3.8 and 3.6, respectively), and we needed the `ioutil` package to be installed in both when running in the container, as it was using the platform python.

With the `venv` they all point to the same version (e.g.):
```
lrwxrwxrwx. 1 root root    7 Feb 15 08:12 python -> python3
lrwxrwxrwx. 1 root root   16 Feb 15 08:12 python3 -> /usr/bin/python3
lrwxrwxrwx. 1 root root    7 Feb 15 08:12 python3.11 -> python3
```
So we always could run this locally, problem was with only for the containers.

nirs:
In the commit message you wrote:

> The platform build was introduced in the commit 
> https://github.com/oVirt/ovirt-imageio/commit/b76751afaf3e260b638a501458be3d61912dfa6e.
> Since platform python (/usr/bin/python3) and
> "python" were referring to different Python
> versions, we needed to build ioutil for both
> to run tests.
>
> Currently this is not required anymore, as
> we run tests inside the virtual environment,
> and all links point to the same version.

I think the issue that we tried to fix was that some tests
use /usr/bin/python3, and some use python, and they may be
different, so we built the ioutil extentsion for both pythons.

This may not be relvant now in the CI, since we split the python
version tests to different jobs, so every job runs with single python
version. But if we are still using /usr/bin/python3, these tests may
fail when running inside the virtual environment.

The other issue is local build, when we *don't* split the build, and
may run the tests with multiple python versions on the same host.
Please confirm that local build still works, it is not enough to have
working CI if you cannot run the tests locally. To test this you need
to install python 3.8 or 3.9 in addition to the platform python
(3.11 in Fedora 37).

aesteve-rh:
I guess the reason for tests using `/usr/bin/python3` would be the shebang?

```
rg 'usr/bin/python3'
ovirt-imageio
1:#!/usr/bin/python3

ovirt-imageioctl
1:#!/usr/bin/python3

ovirt-img
1:#!/usr/bin/python3

ovirt-imageio.spec.in
38:%define __python3 /usr/bin/python3.9
41:%define __python3 /usr/bin/python3
49:%define __python3 /usr/bin/python3.9
52:%define __python3 /usr/bin/python3
180:%define __python3 /usr/bin/python3.9

examples/imageio-client
1:#!/usr/bin/python3

examples/sparse-stream
1:#!/usr/bin/python3

examples/checksum
1:#!/usr/bin/python3

demos/upload-from-ova
1:#!/usr/bin/python3

examples/libvirt-stream
1:#!/usr/bin/python3

test/nbd-client
1:#!/usr/bin/python3

build-aux/release
1:#!/usr/bin/python3
```

If these are a problem, we could use `#!/usr/bin/env python3` instead for better portability.

nirs:
> I guess the reason for tests using `/usr/bin/python3` would be the shebang?

I don't think we run an of these tools in the tests.

I think we had some tests starting the server, and these may use /usr/bin/python3.
based on your search we don't have sch tests now.

If we don't have any code running /usr/bin/python3 in the tests, and we don't call
any of the scripts you fond in the tests, I don't see a reason to built ioutil 
twice, but to very this you need to:

1. install more than one python versions locally
   (e.g. `dnf install python3.9`
2. run tox locally - it will try to test python 3.11 and 3.9, the tests will pass
   if ioutil is not built for both before running the tests

aesteve-rh:
Tested locally, installed Python3.6 in a Fedora37 container with Python 3.11 installed. Run all expected tests environments (`reuse`, `flake8`, `test-py36`, `test-py311`, `bench-py36`, `bench-py311`), and none failed/skipped due to a missing `ioutil`.

These were the only tests skipped in the logs:
```
SKIPPED [1] test/backup_test.py:68: /var/tmp/imageio-images/cirros-0.3.5.img not available
SKIPPED [1] test/backup_test.py:103: /var/tmp/imageio-images/cirros-0.3.5.img not available
SKIPPED [1] test/server_test.py:146: Requires unprivileged user
SKIPPED [1] test/uhttp_test.py:71: Not compatible when running with root
```

nirs:
> Tested locally, installed Python3.6 in a Fedora37 container with Python 3.11 installed. Run all expected tests environments (`reuse`, `flake8`, `test-py36`, `test-py311`, `bench-py36`, `bench-py311`), and none failed/skipped due to a missing `ioutil`.

Sounds good

> SKIPPED [1] test/backup_test.py:68: /var/tmp/imageio-images/cirros-0.3.5.img not available
> SKIPPED [1] test/backup_test.py:103: /var/tmp/imageio-images/cirros-0.3.5.img not available

I think the cirros image is created now when running make, but this test does not
run any child process so it is not related.

> SKIPPED [1] test/server_test.py:146: Requires unprivileged user
> SKIPPED [1] test/uhttp_test.py:71: Not compatible when running with root

These are the tests that may be related, because we support imageio server
only on platform python, and we may run the tests with another python version.
Maybe this is not relevant now since we removed the dependency on systemd
daemon package, which is available only on platform python, so now imageio
server can run with any python version.

Can you run the tests as unprivileged user?

The best way to test this is to run the tests directly on your machine, not in
a container. This is the expected use case, running the tests directly from your
editor.

aesteve-rh:
These are the tests skipped with an unprivileged user, no errors with missing `ioutil`.
```
SKIPPED [1] test/backup_test.py:68: /var/tmp/imageio-images/cirros-0.3.5.img not available
SKIPPED [1] test/backup_test.py:103: /var/tmp/imageio-images/cirros-0.3.5.img not available
SKIPPED [1] test/ioutil_test.py:37: Requires root
SKIPPED [1] test/ioutil_test.py:50: Requires root
SKIPPED [1] test/ioutil_test.py:64: Requires root
SKIPPED [1] test/ioutil_test.py:77: Requires root
SKIPPED [1] test/ioutil_test.py:83: Requires root
SKIPPED [1] test/ioutil_test.py:89: Requires root
SKIPPED [1] test/server_test.py:163: Requires root
SKIPPED [1] test/server_test.py:181: Requires root
```

nirs (approved):
Based on Albert testing this looks safe.

PR #187, opened 2023-03-07
Reorder python builds

nirs:
Based on #186:
- Squash fix commits to one commit
- Add missing signed-off-by

### DIFF
--- a/ovirt-imageio.spec.in
+++ b/ovirt-imageio.spec.in
@@ -41,40 +41,44 @@ Transfer disk images on oVirt system.
 %setup -q
 
 %build
+
+%define python3_pkgversion 3
+%define __python3 /usr/bin/python3
+%py3_build
+
 %if %{with_python39}
 %define python3_pkgversion 39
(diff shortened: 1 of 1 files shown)


nirs:
EPEL9* builds still fail with this change but the rest of the builds are fine.
https://copr.fedorainfracloud.org/coprs/nsoffer/ovirt-imageio-preview/build/5602633/

mwperina:
EPEL9 builds are failing because python 3.11 is not yet there:
```
No matching package to install: 'python3.11-devel'
No matching package to install: 'python3.11-setuptools'
```
So this will probably be fixed only when RHEL 9.2 is released, right?

michalskrivanek (approved):

aesteve-rh (approved):
I'll make a new release then.

michalskrivanek:
releases are fine (CBS builds), this basically just fixes COPR/nightly

aesteve-rh:
Great, thanks @michalskrivanek for the fix and @nirs for putting it together :)

PR #197, opened 2023-03-23
Add Python 3.11 subpackage to be usable in ansible-core 2.14 for el8

mnecas:


nirs (commented):
Ansible is endless fun :-)

Looks good to me, but we probably need to add python 3.11 to the containers?

mnecas:
yeah and it will not stop :( 
https://github.com/oVirt/ovirt-imageio/pull/198

aesteve-rh (approved):

PR #206, opened 2023-04-26
Default values for ovirt-img add_sub_command method added

### LINKED ISSUE #120: Add default values for ovirt-img add_sub_command method
In `test/client_options_test.py` we repeat this a lot for options parse testing:
```python
    parser = _options.Parser()
    parser.add_sub_command(
        "test", "help", lambda x: None, transfer_options=False)
```
Which is actually good since
this is important interface and we want to tests to be easy to read.

However, it would be nice to change the method defaults to make it cleaner
to call this in the tests. So that doing just:
```python
    parser.add_sub_command("test")
```
This will create a command with no help description and a function that does nothing.

sri-srikanth:
ovirt-img add_sub_command method's help and func parameters now have
default values that are set to "help" and a no-op lambda func,
respectively, to simplify method usage in test cases. Method calls
referencing the add_sub_command method have also been cleaned up
accordingly.

Fixes #120 
Signed-off-by: sri-srikanth <srivatsansrikanth2018@gmail.com>

### DIFF
--- a/ovirt_imageio/client/_options.py
+++ b/ovirt_imageio/client/_options.py
@@ -190,7 +190,8 @@ class Parser:
             version=f'%(prog)s {version.string}')
         self._commands = self._parser.add_subparsers(title="commands")
 
-    def add_sub_command(self, name, help, func, transfer_options=True):
+    def add_sub_command(self, name, help="help", func=lambda x: None,
+                        transfer_options=True):
         cmd = self._commands.add_parser(name, help=help)
         cmd.set_defaults(command=func)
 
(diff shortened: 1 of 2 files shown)


aesteve-rh:
Do the reviews over the same commit, to have a clean commit history.

In this case, you can squash the commits with an interactive rebase:
```
git rebase -i HEAD~2
```
And then change `pick` for `squash` on the last commit and that should merge them.
Then, you can force push to update the remote branch (and the PR status).
```
git push -f
```
No need to open a new PR for changes.

If you have any issue, let me know!

sri-srikanth:
Squashed :)

aesteve-rh:
The change looks good, but could you please change the commit text with the one you used for the PR?

Including the `Signed-off` line and the `Fixes #120`.
Try to fit the commit title into 72 characters if posible.

PR #214, opened 2025-04-06
Remove stale copr build tag

nirs:
The copr repo nsoffer/ovirt-imageio-preview is not maintained for few years and does not exist now.

If we want to link to copr report for builds it should be the official oVirt copr repo.

The broken tag:
broken-tag


dupondje:
Thanks
We have a new URL, so we can just replace it with the new one. See PR #216

PR #220, opened 2025-07-08
build: fix builds with new setuptools

dupondje:
Since setuptools v69.3.0, name of the tar file is modified according to PEP 625. Therefor update the code so it builds correctly again with newer setuptools versions.


dupondje:
Builds on COPR work again after this:
https://copr.fedorainfracloud.org/coprs/ovirt/ovirt-master-snapshot/build/9249382/

JasperB-TeamBlue (changes requested):
I feel like it is against best practice to mix ovirt-imageio and ovirt_imageio. Should it not be better if it was written everywhere one of either way?

JasperB-TeamBlue (approved):

